PR #20, opened 2022-01-11
Provision dynamic TLS certificates

### LINKED ISSUE #3: TLS cert provisioning
The operator should (optionally) be able to provision TLS certificates (and keys) dynamically, rather than reading them from static secrets.

Acceptance criteria:

- [ ] Backend/type configuration - #8
- [ ] Load CA certificate
- [ ] Bootstrap CA certificate if missing
- [ ] Create pod certificate for a given scope (node FQDN, pod FQDN, etc) on demand

teozkr:
Fixes #3 

- [X] Factor out secret source interface
- [X] Create TLS source
- [X] Store CA certificate
- [ ] Configurable secret classes
- [X] Docs


teozkr:
As discussed with @soenkeliebau, we're currently punting on configurable secret classes to instead focus on getting a minimal MVP for TLS out that other operators can start using.

soenkeliebau:
I have looked through the code - I won't pretend to understand the GRPc stuff or how the entire thing fits together, so my review will be from a functional point of view.

I would have liked a few more comments in the code, but I am fully aware that I was also the one pushing to get this out the door quickly, so I'll shut up already :)

We are currently using the provisioner in the NiFi and Trino operator rework branches and it works as expected. 

The only thing that I noticed is that the external ip address of the node the pod is assigned to is missing from the cert. It contains the internal ip in the overlay network as well as the hostname, but not the external ip assigned to that host.

This would be a useful addition to allow external communication via nodeport etc to the pod.
image

```
Addresses:
  InternalIP:   10.128.0.16  
  ExternalIP:   104.197.3.74
  InternalDNS:  gke-nifi2-default-pool-3f4b62f3-7n9w.us-central1-c.c.engineering-329019.internal  
  Hostname:     gke-nifi2-default-pool-3f4b62f3-7n9w.us-central1-c.c.engineering-329019.internal  
```

For load balancing use cases we might actually want to consider adding the external ips of all nodes to the cert on request, but that is in the future I'd say.

teozkr:
The node's `InternalIP` should typically be the off-overlay LAN IP, as opposed to the internet-reachable `ExternalIP`. Agreed that the certificate should ideally contain all of the node's IP addresses.

teozkr:
Added secondary node IPs, as discussed.

src/backend/tls.rs
```diff
@@ -0,0 +1,247 @@
...
+        client: &stackable_operator::client::Client,
+    ) -> Self {
+        let k8s_secret_name = "secret-provisioner-ca";
+        let k8s_ns = "default";
+        let existing_secret = client.get::<Secret>(k8s_secret_name, Some(k8s_ns)).await;
+        match existing_secret {
+            Ok(ca_secret) => {
+                let ca_data = ca_secret.data.unwrap_or_default();
+                Self {
+                    ca_key: PKey::private_key_from_pem(&ca_data.get("ca.key").unwrap().0).unwrap(),
```
soenkeliebau, line 117: Not sure what would be the correct behavior here, we probably don't want to throw away anything in the secret and regenerate a ca just because there was a maybe spurious parsing failure.
But at least returning an error would be better?

src/backend/tls.rs
```diff
@@ -0,0 +1,247 @@
...
+                            .name(k8s_secret_name)
+                            .build(),
+                        data: Some(
+                            [
+                                (
+                                    "ca.key".to_string(),
+                                    ByteString(ca.ca_key.private_key_to_pem_pkcs8().unwrap()),
+                                ),
+                                (
+                                    "ca.crt".to_string(),
```
soenkeliebau, line 138: Are these unwraps safe? If so, can we replace them with .expect() ?

src/backend/tls.rs
```diff
@@ -0,0 +1,247 @@
...
+    type Error = Error;
+
+    async fn get_secret_data(
+        &self,
+        selector: super::SecretVolumeSelector,
+        pod_info: PodInfo,
+    ) -> Result<SecretFiles, Self::Error> {
+        let now = OffsetDateTime::now_utc();
+        let not_before = now - Duration::minutes(5);
+        let not_after = now + Duration::days(1);
```
soenkeliebau, line 165: I think this should be significantly higher, as most of the services we provision will run for a long time without a restart and do not perform dynamic reloading of certificates.
Ideally this should be configurable by the requestor at some point.

soenkeliebau (commented):
I have left a few comments, mostly minor stuff.

There are a few things in here that I think we should like to make configurable, but that is something that can be addressed in the future. 
This is an excellent first implementation that does what we currently need it to do, thanks!

README.md
```diff
@@ -0,0 +1,55 @@
...
+
+The operator injects secret data into `Pod` `Volume`s that declare a CSI volume with the `driver` `secrets.stackable.tech`.
+The volume takes the following mandatory attributes:
+
+- `secrets.stackable.tech/class`: The type of secret to be issued. Supported options:
+  - `tls`
+- `secrets.stackable.tech/scope`: The properties of the `Pod` that the secret should authenticate. Supported options:
+  - `node`
+  - `pod`
+  - `secret=<foo>`
```
soenkeliebau, line 29:
```suggestion
  - `service=<foo>`
```

src/backend/tls.rs
teozkr: Hence why we `create()` the `Secret` rather than `apply()`ing it. If it already exists then that should return with a conflict failure, which we present to the user. We might want to add a logging statement or something here though, to help understand why the creation path is being invoked in the first place?

src/backend/tls.rs
teozkr: They are safe in the sense that they don't depend on any user-provided data.

I agree that `expect` might be clearer, but #21 turns them into regular `Error`s (since we start pulling in the data from the `SecretClass` rather than hard-coding it).

src/backend/tls.rs
teozkr: We'll need to handle reloading the certificates somehow anyway (whether that be through hot reloading or cold-blooded murder), regardless of whether we make the expiry time 5 minutes or 5 weeks. Agreed that we'll probably need to handle long-lasting certificates *somehow*, but I think that should be the exception rather than the rule.

src/backend/tls.rs
soenkeliebau: Just noticed that some of my comments were addressed in #21 yes, came back here to mention that, but you were too quick for me!

src/backend/tls.rs
soenkeliebau: For the large majority of products I think a restart is required to activate new certificates, which can make this an expensive operation. But wherever possible we should make a hot-reload possible.

To mitigate this, could we make the runtime user configurable? Maybe the SecretClass for AutoTls prvisioners could get settings for `defaultCertValidity` and `maximumCertValidity`?
If nothing is specified in the volume default gets pulled and within the other parameter longer validities can be requested by a parameter on the volume.

src/backend/tls.rs
soenkeliebau: My comment was mostly referring to the unwraps that might fail if for example "ca.key" was mistyped in the Secret.

But these have been fixed in #21 so you can disregard this comment :)

src/backend/tls.rs
teozkr: `maxCertValidity` on the `SecretClass` makes sense to me. `defaultCertValidity` feels like a footgun that should be managed by the individual operators instead (I'm a bit terrified of the future medium/stackoverflow answers telling people that they should "just set `defaultCertValidity` to a year or whatever" :P).

soenkeliebau: That works for me.

soenkeliebau (approved):

PR #21, opened 2022-01-14
Configurable SecretClasses

teozkr:
This builds upon #20, but should be merged into `main` as usual. I selected that as the target branch for ease of review.

This adds a `SecretClass` CRD intended to be configured by the cluster operator, which contains policy information about how secrets should be issued. For example, this is where the administrator would configure whether to use automatically issued TLS certificates, or ones issued manually by the administrator.


Cargo.toml
```diff
@@ -33,4 +35,11 @@ tonic-build = "0.6.0"
 # Workaround for https://github.com/hyperium/tonic/issues/243
 h2 = { git = "https://github.com/teozkr/h2.git", branch = "feature/grpc-uds" }
 
+kube = { git = "https://github.com/teozkr/kube-rs.git", branch = "feature/crd-complex-enum" }
```
soenkeliebau, line 38: I think this can be removed now?

README.md
```diff
@@ -53,3 +52,20 @@ spec:
     - name: tls
       mountPath: /tls
 ```
+
+`SecretClass` defines where the secrets come from. For example, the following `SecretClass`
```
soenkeliebau, line 56: Can we add a list of currently available implementations and the config parameters?

soenkeliebau (commented):
I've looked through the code and am happy in principle. Left two comments, mainly around docs.
I've not tested this yet and have held back on approving since it'll need to be rebased on main after merging #20 anyway.

Cargo.toml
teozkr: Yes, once we upgrade to operator-rs 0.8.0.

README.md
teozkr: Done, but we'll probably want to reformat it at some point.

soenkeliebau (approved):

PR #22, opened 2022-01-17
Added a few comments.

soenkeliebau:
I've tried my hand at adding a few high level comments.


src/main.rs
```diff
@@ -223,6 +246,10 @@ impl Node for SecretProvisionerNode {
         Ok(Response::new(NodePublishVolumeResponse {}))
     }
 
+    // Called when a node is terminated that contained a volume created by this provider.
```
teozkr, line 249:
```suggestion
    // Called when a pod is terminated that contained a volume created by this provider.
```

PR #26, opened 2022-01-26
Lock down file permissions on CSI socket

### LINKED ISSUE #16: Secure file permissions on CSI socket

teozkr:
Fixes #16

This is a bit more complicated than it ought to be, because neither Tokio nor std expose a proper API for this (see https://github.com/tokio-rs/tokio/issues/4422).


teozkr:
bors try

maltesander (approved):
LGTM.

PR #27, opened 2022-01-26
[Merged by Bors] - Helm chart

### LINKED ISSUE #25: Helm packaging

teozkr:
Fixes #25 

For now I applied the templating manually (from https://github.com/stackabletech/operator-templating/pull/55, commit https://github.com/stackabletech/operator-templating/pull/55/commits/4fd238d4af9b59cf0c771b53799e22bf11ac87db) and adapted it for secret-operator by hand.

These changes are contained in 878c4a6176782450c068669638c92700ebcb65d7, and basically mean:
- Deploying as a `DaemonSet` rather than a `Deployment`
- Adding CSI sidecar containers
- Running as root (at some point we might be able to get away with "just" adding a bunch of capabilities, but secret-operator will probably always require relatively elevated privileges)
- Disabling product-config
- Extracting the CRD by running `cargo run crd` rather than reading the file that `build.rs` creates (Nix doesn't allow access to files outside of the source tree)
- Adapting to the `ClusterRole` rules that secret-operator requires

Currently this branch contains both the `Dockerfile`- and Nix-based workflows for building the operator, but we should probably pick one at some point.


.github/workflows/daily_security.yml
```diff
@@ -2,11 +2,12 @@
 # This file is automatically generated from the templates in stackabletech/operator-templating
 # DON'T MANUALLY EDIT THIS FILE
 # =============
+---
 name: Security audit
 
 on:
   schedule:
-    - cron: '0 0 * * *'
+    - cron: '15 4 * * *'
```
maltesander, line 10: whats the reason for this? to have it run before the '30 4 * ...' stuff?

deploy/helm/secret-operator/README.md
```diff
@@ -0,0 +1,33 @@
+# Helm Chart for Stackable Operator for Stackable Secret Operator
+
+This Helm Chart can be used to install Custom Resource Definitions and the Operator for Stackable Secret Operator provided by Stackable.
```
maltesander, line 3: Ok i dont think that naming makes sense, but its consistent :D
`This Helm Chart can be used to install Custom Resource Definitions and the Secret Operator provided by Stackable.`

deny.toml
```diff
@@ -56,4 +53,4 @@ unknown-registry = "deny"
 unknown-git = "deny"
 
 [sources.allow-org]
-github = ["stackabletech", "teozkr"]
\ No newline at end of file
+github = ["stackabletech", "teozkr"]
```
maltesander, line 56: Do we get anything from your repo here still?

maltesander (approved):
Just a question and minor labeling (at least for readme, not in the generated stuff).
LGTM otherwise.

.github/workflows/daily_security.yml
teozkr: Not sure, this came from the templating

deny.toml
teozkr: The patched unix domain stream-friendly h2 build.

PR #64, opened 2022-02-25
[Merged by Bors] - Docs

### LINKED ISSUE #12: Docs
As a user of or an admin running the Stackable platform I need to know that this operator exists, what it does and how I can use it.
As a developer of the platform I need documentation on how to utilize the operator in my own operator.

This is done when we have documentation in the contributors guide and the regular documentation on docs.stackable.tech.

teozkr:
## Description

Fixes #12. Not super happy about this, but hopefully it's an improvement over the current state, at least...

## Review Checklist
- [ ] (Integration-)Test cases added (or not applicable)
- [ ] Documentation added (or not applicable)
- [ ] Changelog updated (or not applicable)

Once the review is done, comment `bors r+` (or `bors merge`) to merge. [Further information](https://bors.tech/documentation/getting-started/#reviewing-pull-requests)


docs/modules/ROOT/pages/secretclass.adoc
```diff
@@ -0,0 +1,117 @@
...
+    # or... <1>
+    k8sSearch: # <3>
+      searchNamespace:
+        pod: {}
+        # or...
+        name: my-namespace
+      secretLabels:
+        type: custom-secret
+----
+<1> Backends are mutually exclusive, only one may be used by each  SecretClass`
```
maltesander, line 33:
```suggestion
<1> Backends are mutually exclusive, only one may be used by each `SecretClass`
```

maltesander (commented):
One fix, lgtm otherwise.

docs/modules/ROOT/pages/usage.adoc
```diff
@@ -0,0 +1,47 @@
...
+      volumeAttributes:
+        secrets.stackable.tech/class: tls
+        secrets.stackable.tech/scope: node,pod,service=secret-consumer-nginx
+  containers:
+  - name: ubuntu
+    image: ubuntu
+    stdin: true
+    tty: true
+    volumeMounts:
+    - name: tls
```
lfrancke, line 27: Does this need to be `secret` instead of `tls`?

docs/modules/ROOT/pages/usage.adoc
```diff
@@ -0,0 +1,47 @@
...
+      mountPath: /tls
+----
+
+xref:secretclass.adoc[] defines where the secrets come from. For example, the following `SecretClass`
+issues TLS certificates, storing its CA certificate in `secret-provisioner-tls-ca`:
```
lfrancke, line 32: I might be dumb but it might be worthwhile saying that it's a Kubernetes `Secret` that it's being stored in. At least I assume so.

docs/modules/ROOT/pages/secretclass.adoc
```diff
@@ -0,0 +1,117 @@
...
+    autoTls:
+      ca:
+        secret:
+          name: secret-provisioner-tls-ca
+          namespace: default
+----
+
+`autoTls`:: Declares that the `autoTls` backend is used.
+`autoTls.ca`:: Configures the certificate authority used to issue `Pod` certificates.
+`autoTls.ca.secret`:: Reference (`name` and `namespace`) to a K8s `Secret` object where the CA certificate and key is stored as `ca.crt`
```
lfrancke, line 66: Do I understand this correctly: It'll create self-signed certificates if the referenced CA doesn't exist?
This seems like a "dangerous" default and super annoying to debug if someone just has a simple typo in there and expects to reference an existing one.

This comment is not about the docs, I know. And if it documents current behavior we should document it, but if my assumption is true then we might want a follow-up ticket.

docs/modules/ROOT/pages/secretclass.adoc
```diff
@@ -0,0 +1,117 @@
...
+
+[#backend-k8ssearch]
+=== `k8sSearch`
+
+*Format*: Free-form
+
+A Kubernetes `Secret` object is selected based on the xref:scope.adoc[scopes] specified on the `Volume`. Each field in this `Secret` is mapped
+to one file. It is suggested these `Secret` objects should follow one of the xref:#format[formats defined in this document].
+
+xref:scope.adoc[Scopes] are translated into additional label filters of the form `secrets.stackable.tech/$SCOPE: $SCOPE_VALUE`.
```
lfrancke, line 77: I read this twice and I still don't understand what it does really.
I think a more "complete" example might help.
One showing a Secret, a Pod and a SecretClass

docs/modules/ROOT/pages/scope.adoc
```diff
@@ -0,0 +1,26 @@
+= Scope
+
+Secrets often cover some specific aspect of a workload. For example:
+
+* A Kerberos credential may be bound to one node IP
+* An internal TLS certificate's `subjectAlternateName` section must match the `Pod`'s name and service
```
lfrancke, line 6: Something's off with the formatting here. 

The  's after Pod seems to not trigger the end of the fixed-width formatting.
You might have to use two backticks, I'm not 100% sure.

teozkr: Good point. I thought I was strict with using "`foo` object's" (awkward as that is, too..), but it looks like I missed one case.

docs/modules/ROOT/pages/secretclass.adoc
teozkr: If the CA doesn't exist then it will create a self-signed CA the first time a certificate is issued, and save that into the k8s `Secret` object that you have specified.

The certificate issued to the pod will never be self-signed.

docs/modules/ROOT/pages/usage.adoc
teozkr: Good point.

docs/modules/ROOT/pages/usage.adoc
teozkr: Good point, this should be changed to `tls` in the volume definition.

docs/modules/ROOT/pages/secretclass.adoc
```diff
@@ -0,0 +1,117 @@
...
+
+This is intended to provide an abstraction between how the secret is used ("I need a certificate for my cluster's TLS PKI") and how
+it is provisioned (automatically and generated by the operator's internal CA, provisioned by the cluster administrator, or provisioned by an
+external service such as Hashicorp Vault).
+
+A `SecretClass` looks like this:
+
+[source,yaml]
+----
+apiVersion: secrets.stackable.tech/v1alpha1
```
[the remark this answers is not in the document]
teozkr: We should probably include these from example files, so that yamllint can pick them up as usual.

docs/modules/ROOT/pages/installation.adoc
```diff
@@ -0,0 +1,29 @@
+= Installation
```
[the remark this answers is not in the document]
teozkr: Yeah.. the "how to build from source" page covers installation as well, but it could certainly be integrated better.

docs/modules/ROOT/pages/secretclass.adoc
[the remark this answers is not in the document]
lfrancke: That's a good idea, yes.

docs/modules/ROOT/pages/secretclass.adoc
lfrancke: Understood, yes. Sorry, I meant CA not certificate.

This is something worth changing IMO.

teozkr: Feels like there is some misalignment here around what the expected use cases would be. Happy to have a discussion here when you're available.

lfrancke (commented):
The changes look good, thank you!

soenkeliebau (approved):

PR #114, opened 2022-04-12
[Merged by Bors] - Mark pods for expiry when secrets are no longer applicable

### LINKED ISSUE #92: Allow pods to opt into getting killed when their secrets are about to expire

teozkr:
## Description

Fixes #92.

- [X] Mark `autoTls` pods for expiry after a set time
- [ ] Mark `k8sSearch` pods for expiry when listed objects are modified - postponed as #118 
- [x] Implement above behaviour in commons operator - https://github.com/stackabletech/commons-operator/pull/20

## Review Checklist
- [ ] (Integration-)Test cases added (or not applicable)
- [ ] Documentation added (or not applicable)
- [ ] Changelog updated (or not applicable)
- [x] Cargo.toml only contains references to git tags (not specific commits or branches)
- [ ] Helm chart can be installed and deployed operator works (or not applicable)

Once the review is done, comment `bors r+` (or `bors merge`) to merge. [Further information](https://bors.tech/documentation/getting-started/#reviewing-pull-requests)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,7 @@ All notable changes to this project will be documented in this file.
 
 ### Added
 
-- Pods that use autoTls volumes are now deleted when their certificates are about to expire ([#114], [commons-#20]).
+- Pods that use autoTls volumes are now evicted when their certificates are about to expire ([#114], [commons-#20]).
 
 ### Changed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,12 +4,18 @@ All notable changes to this project will be documented in this file.
 
 ## [Unreleased]
 
+### Added
+
+- Pods that use autoTls volumes are now deleted when their certificates are about to expire ([#114], [commons-#20]).
+
 ### Changed
 
 - autoTls CA generation now requires opt-in ([#77]).
   - The default `tls` `SecretClass` now has this opt-in by default.
 
 [#77]: https://github.com/stackabletech/secret-operator/pull/77
+[#114]: https://github.com/stackabletech/secret-operator/pull/114
+[commons-#20]: https://github.com/stackabletech/commons-operator/pull/20
 
 ## [0.2.0] - 2022-02-14
 
--- a/rust/operator-binary/src/backend/mod.rs
+++ b/rust/operator-binary/src/backend/mod.rs
@@ -8,6 +8,7 @@ pub mod tls;
 
 use async_trait::async_trait;
 use serde::Deserialize;
+use stackable_operator::k8s_openapi::chrono::{DateTime, FixedOffset};
 use std::{collections::HashMap, convert::Infallible, path::PathBuf};
 
 pub use dynamic::Dynamic;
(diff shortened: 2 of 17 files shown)


teozkr:
Do we want this to be opt-in (but disabled by default), opt-out (but enabled by default), or mandatory? Originally the issue was written assuming that this would be opt-in, but thinking about it a bit more I'm leaning towards making it mandatory for now, and transitioning towards opt-out once we implement hot-updating secrets in secret-op.

rust/operator-binary/src/backend/mod.rs
```diff
@@ -84,6 +85,24 @@ impl SecretVolumeSelector {
 }
 
 type SecretFiles = HashMap<PathBuf, Vec<u8>>;
+#[derive(Default, Debug)]
```
sbernauer, line 88: NIT
```suggestion

#[derive(Default, Debug)]
```

rust/operator-binary/src/backend/mod.rs
```diff
@@ -84,6 +85,24 @@ impl SecretVolumeSelector {
...
+pub struct SecretContents {
+    pub files: SecretFiles,
+    pub expires_after: Option<DateTime<FixedOffset>>,
+}
+impl SecretContents {
```
sbernauer, line 93: Nit
```suggestion

impl SecretContents {
```

rust/operator-binary/src/backend/tls.rs
```diff
@@ -307,26 +311,34 @@ impl SecretBackend for TlsGenerate {
...
+                        .context(SerializeCertificateSnafu { tpe: CertType::Pod })?,
+                ),
+            ]
+            .into(),
+        )
+        .expires_after(time_datetime_to_chrono(expire_pod_after)))
     }
 }
+
+fn time_datetime_to_chrono(dt: time::OffsetDateTime) -> chrono::DateTime<FixedOffset> {
```
sbernauer, line 341: Would you mind adding a test for this function? I personally find the conversion not straight forward, on small example conversion as test-case would help

sbernauer (commented):
Looks good overall :+1: 

Could you please add a Changelog entry?

I'm asking myself if having a hash as tag is a good idea of if it would be better to use the volume name as the tag because of the following reasons:
- Easier to implement
- Easier to understand for the user (descriptive names) -> They could also set up monitoring based on the descriptive Pod Labels (I'm used to having that metrics in Prometheus already)
- No hash collisions

The resulting label names should be always valid as the volume names are also enforced to the same extend as the name below AFAIK
> Annotations are key/value pairs. Valid annotation keys have two segments: an optional prefix and name, separated by a slash (/). The name segment is required and must be 63 characters or less, beginning and ending with an alphanumeric character ([a-z0-9A-Z]) with dashes (-), underscores (_), dots (.), and alphanumerics between

teozkr:
> Easier to understand for the user (descriptive names) -> They could also set up monitoring based on the descriptive Pod Labels (I'm used to having that metrics in Prometheus already)

Yeah, this is one reason _for_ the hash IMO, since this is more of an implementation detail than a part of secret-op's public API.

We can always change this to something more human-readable later on, if we want to..

> No hash collisions

On the other hand, we increase the risk of name collisions, since we can't really namespace them anymore (without the annotations becoming gargantuan, anyway).

>  Do you think it is overkill to check if we have a hash collision by reading the existing Pod and make sure we don't have a collision (and if so use another tag)?

Well, there _can_ be valid reasons to overwrite the annotation, if we're retrying a previously failed provisioning (or reprovisioning the secret for an existing pod). If we want to tackle this more seriously then I'd rather switch to a "real" cryptographic hash (like SHA-2 or Blake2...).

sbernauer (approved):
Please add it to the CHANGELOG and were good to go. Thanks!

teozkr:
@sbernauer Added changelog and replaced fnv with sha2.

CHANGELOG.md
```diff
@@ -4,12 +4,18 @@ All notable changes to this project will be documented in this file.
 
 ## [Unreleased]
 
+### Added
+
+- Pods that use autoTls volumes are now deleted when their certificates are about to expire ([#114], [commons-#20]).
```
sbernauer, line 9:
```suggestion
- Pods that use autoTls volumes are now evicted when their certificates are about to expire ([#114], [commons-#20]).
```

sbernauer (approved):

teozkr:
bors retry

PR #304, opened 2023-08-10
More docs on the operator's added value

razvan:


sbernauer (approved):

adwk67 (approved):

fhennig:
It might be also good to mention that the operator can rotate the secrets - as opposed to the static default secrets.

PR #311, opened 2023-09-01
Use eval instead of yq in the integration tests to substitute the namespace

siegfriedweber:
# Description

Use eval instead of yq in the integration tests to substitute the namespace.

yq is not installed on the test driver and the integration tests of all other operators use eval to substitute the namespace, so it makes sense to do the same for the secret-operator.

```[tasklist]
# Author
```

```[tasklist]
# Reviewer
- [ ] Documentation added or updated
- [ ] Changelog updated
```

```[tasklist]
# Acceptance
```

### DIFF
--- a/tests/templates/kuttl/kerberos-ad/01-install-secretclass.yaml
+++ b/tests/templates/kuttl/kerberos-ad/01-install-secretclass.yaml
@@ -2,7 +2,7 @@
 apiVersion: kuttl.dev/v1beta1
 kind: TestStep
 commands:
-  - script: yq eval '.metadata.name = "kerberos-" + strenv(NAMESPACE) | .spec.backend.kerberosKeytab |= (.adminKeytabSecret.namespace = strenv(NAMESPACE) | .admin.activeDirectory.passwordCacheSecret.namespace = .adminKeytabSecret.namespace | .admin.activeDirectory.ldapTlsCaSecret.namespace = .adminKeytabSecret.namespace)' secretclass.yaml | kubectl apply -f-
+  - script: eval "echo \"$(cat secretclass.yaml)\"" | kubectl apply -f -
 ---
 apiVersion: v1
 kind: Secret
--- a/tests/templates/kuttl/kerberos-ad/secretclass.yaml
+++ b/tests/templates/kuttl/kerberos-ad/secretclass.yaml
@@ -1,9 +1,7 @@
-# must be applied by a command step, since the reference namespace depends on the (random) test namespace
 ---
 apiVersion: secrets.stackable.tech/v1alpha1
 kind: SecretClass
 metadata:
-  # overridden by yq
   name: kerberos-$NAMESPACE
 spec:
   backend:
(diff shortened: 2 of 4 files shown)


razvan (approved):
nice. much cleaner.

tests/templates/kuttl/kerberos-ad/01-install-secretclass.yaml
```diff
@@ -2,7 +2,7 @@
 apiVersion: kuttl.dev/v1beta1
 kind: TestStep
 commands:
-  - script: yq eval '.metadata.name = "kerberos-" + strenv(NAMESPACE) | .spec.backend.kerberosKeytab |= (.adminKeytabSecret.namespace = strenv(NAMESPACE) | .admin.activeDirectory.passwordCacheSecret.namespace = .adminKeytabSecret.namespace | .admin.activeDirectory.ldapTlsCaSecret.namespace = .adminKeytabSecret.namespace)' secretclass.yaml | kubectl apply -f-
+  - script: eval "echo \"$(cat secretclass.yaml)\"" | kubectl apply -f -
```
nightkr, line 5: If I understand this right, the shell substitution would happen in the outer shell, right? Wouldn't that mean that this breaks as soon as the YAML contains a double quote?

tests/templates/kuttl/kerberos-ad/kinit-client.yaml
```diff
@@ -21,11 +22,11 @@ spec:
               echo listing keytab contents
               klist -k /stackable/krb/keytab -teKC
               echo kiniting node
-              kinit -kt /stackable/krb/keytab -p HTTP/$NODE_NAME
+              kinit -kt /stackable/krb/keytab -p HTTP/\$NODE_NAME
```
nightkr, line 25: This also seems likely to trip someone up who is looking at the test files as an example.

nightkr (changes requested):

siegfriedweber:
https://ci.stackable.tech/view/02%20Operator%20Tests%20(custom)/job/secret-operator-it-custom/4/

tests/templates/kuttl/kerberos-ad/01-install-secretclass.yaml
siegfriedweber: > If I understand this right, the shell substitution would happen in the outer shell, right?

Yes, in the shell started by KUTTL.

> Wouldn't that mean that this breaks as soon as the YAML contains a double quote?

Double quotes are removed. If they are required then they must be escaped:

```
$ cat test.yaml
unquotedNamespace: $NAMESPACE.suffix
quotedNamespace: "$NAMESPACE.suffix"
requiredQuotes: \"1\"

$ NAMESPACE=mynamespace sh -c 'eval "echo \"$(cat test.yaml)\""'
unquotedNamespace: mynamespace.suffix
quotedNamespace: mynamespace.suffix
requiredQuotes: "1"
```

tests/templates/kuttl/kerberos-ad/kinit-client.yaml
siegfriedweber: True. Strictly speaking, this is not a YAML file but a templated file meant to be evaluated by a shell. So the file name should actually be suffixed by something, e.g. `kinit-client.yaml.with-envs`, to make it clear that it must be preprocessed before it can be applied.

We have to discuss it in our team if we want to add `yq` to the test driver or if we want to use `eval` or `kubectl apply -f - <<EOF` or anything else.

siegfriedweber:
with `envsubst` instead of `eval`: https://ci.stackable.tech/view/02%20Operator%20Tests%20(custom)/job/secret-operator-it-custom/5/

nightkr:
That Jenkins URL 404s for me: 
image

nightkr:
Ah, I guess the test details are now private.

nightkr (approved):
LGTM

PR #314, opened 2023-09-06
Add support for encrypting PKCS#12 stores with a password

### LINKED ISSUE #313: Allow users to override PKCS#12 password
Some applications (HDFS, grumble grumble) refuse to load certificates from keystores with no password.

nightkr:
# Description

Fixes #313.

We add a new volume parameter `secrets.stackable.tech/format.compatibility.tls-pkcs12.password` for controlling this. The password should not be considered private (PVC manifests are generally public!), this is purely for compatibility with *certain* broken apps.

```[tasklist]
# Author
- [ ] CRD changes approved
```

```[tasklist]
# Reviewer
- [x] Changelog updated
- [x] Cargo.toml only contains references to git tags (not specific commits or branches)
```

```[tasklist]
# Acceptance
```


nightkr:
I'm not sure if this is something that we want to document, since this is more of a desperate "feature of last resort".

rust/operator-binary/src/backend/mod.rs
```diff
@@ -72,6 +72,17 @@ pub struct SecretVolumeSelector {
         deserialize_with = "SecretVolumeSelector::deserialize_str_vec"
     )]
     pub kerberos_service_names: Vec<String>,
+
+    /// A password that the TLS PKCS#12 keystore should be protected using
```
maltesander, line 76:
```suggestion
    /// A password to protect the generated TLS PKCS#12 keystores
```

maltesander (dismissed):
LGTM! Lets wait what the arch meeting brings.

rust/operator-binary/src/backend/mod.rs
nightkr: On second thought, I don't like "protect" at all in this case. We're not protecting anything, just adding yet another box ticking exercise.

maltesander: Yeah im fine with another wording.

siegfriedweber: In my opinion, this sentence is grammatically incorrect. We should fix it before merging. What about "The passphrase used to encrypt the TLS PKCS#12 keystore"?

siegfriedweber (changes requested):
In general okay after the comment for `compat_tls_pkcs12_password` is fixed.

I would not emphasize it so much that the password is used for compatibility reasons. It is an option (and not a compatibility option) which is used for compatibility and not too useful on its own.

rust/operator-binary/src/backend/mod.rs
nightkr: be3ab3ce2675b137512394d8da9d8217c9316f33

I stuck to password for consistency.

nightkr:
It *only* exists for compatibility, and provides no useful security improvement on its own.

I absolutely want to avoid misleading people around this...

siegfriedweber (approved):
LGTM

sbernauer (approved):

PR #324, opened 2023-11-08
Make tests succeed on OpenShift

razvan:
Tested on OS 14.12.25

:green_circle:  https://ci.stackable.tech/view/02%20Operator%20Tests%20(custom)/job/secret-operator-it-custom/8/


adwk67 (approved):
LGTM.

PR #343, opened 2024-01-10
Add shell.nix for local development environments

NickLarsenNZ:
# Description

> [!NOTE]
> I considered adding this to operator-templating, but didn't because:
> - Some operators have different dependencies than listed in `shell.nix` and it doesn't seem like a good idea to manage dependencies over two repos. Consider each operator independently.
> - This is the first, and it might be worth trying it out in a few operators before deciding if we want it templated.

- `shell.nix` (for use with `nix-shell or nix-shell --pure`), or autoloaded...
- `.envrc` for use with [direnv] and [nix-direnv] (run `direnv allow` to allow direnv to load the shell.nix automatically)

[direnv]: https://direnv.net/
[nix-direnv]: https://github.com/nix-community/nix-direnv

```[tasklist]
# Reviewer
- [ ] Documentation added or updated. Follows the [style guide](https://docs.stackable.tech/home/nightly/contributor/style-guide).
- [ ] Changelog updated
```


NickLarsenNZ:
Closed because I checked out from another branch and didn't want to keep the mess after fixing it.

PR #344, opened 2024-01-10
chore: add shell.nix for local dev environments

NickLarsenNZ:
# Description

> [!NOTE]
> I considered adding this to operator-templating, but didn't because:
> - Some operators have different dependencies than listed in `shell.nix` and it doesn't seem like a good idea to manage dependencies over two repos. Consider each operator independently.
> - This is the first, and it might be worth trying it out in a few operators before deciding if we want it templated.

- `shell.nix` (for use with `nix-shell` or `nix-shell --pure`), or autoloaded...
- `.envrc` (sample) for use with [direnv] and [nix-direnv] (run `direnv allow` to allow direnv to load the shell.nix automatically)

[direnv]: https://direnv.net/
[nix-direnv]: https://github.com/nix-community/nix-direnv

```[tasklist]
# Reviewer
- [ ] Documentation added or updated. Follows the [style guide](https://docs.stackable.tech/home/nightly/contributor/style-guide).
- [ ] Changelog updated
```


shell.nix
```diff
@@ -0,0 +1,45 @@
...
+  buildInputs = with pkgs; [
+    # cargo et-al
+    rustup
+
+    # make regenerate charts
+    yq-go
+    git
+    jq
+
+    # make run-dev
```
nightkr, line 20: make run-dev should pull in its dependencies via nix already, we don't need them in shell.nix

shell.nix
```diff
@@ -0,0 +1,45 @@
+let 
+  pkgs = import <nixpkgs> {};
```
nightkr, line 2: This should use the pinned (niv) nixpkgs instead.

shell.nix
```diff
@@ -0,0 +1,45 @@
...
+  beku = pkgs.callPackage(pkgs.fetchFromGitHub {
```
nightkr, line 3: beku should also be pinnable with niv, so we handle it all in one place.

shell.nix
NickLarsenNZ: I was getting errors about missing binaries when I ran `make run-dev` without the listed dependencies. I tested in `nix-shell --pure` - does that work for you without the listed deps?

_Btw, you can select multiple consecutive lines to comment on so that there is more context in the comment view._

shell.nix
```diff
@@ -0,0 +1,45 @@
...
+    # tests
+    kuttl
+    beku
+
+    pkg-config
+    clang
+    libclang
+    openssl
+    protobuf
+    krb5
```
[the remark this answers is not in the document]
NickLarsenNZ: That would be ideal, though their names can differ (unless there's some fancy way to handle that?)

shell.nix
NickLarsenNZ: Is that a change to make in the beku.py repo? Or here?

shell.nix
NickLarsenNZ: What change do you suggest here? I'm not familiar with niv.

shell.nix
nightkr: I know.. I was on mobile so I think the selection got screwed up.. :(

shell.nix
nightkr: Here. `niv add stackabletech/beku.py`, then grab it from `nix/sources.nix`.

shell.nix
NickLarsenNZ: Ah yeah.

Can this be considered resolved, given my answer? Or have I neglected something?

shell.nix
NickLarsenNZ: Ah ok cool, I think I understand.
Will ping you once I think I've done it.

shell.nix
nightkr: `nix-shell --pure` fails to even evaluate for me:

```
n@emeritus:~/Documents/stackable/secret-provisioner$ nix-shell --pure
building '/nix/store/60prdsmj7p46jjz25dj2zr5iy7fgk377-source.drv'...

trying https://github.com/stackabletech/beku.py/archive/589e48ae45c5984d8f1528e1c0b802e9fa137715.tar.gz
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0
100 22970    0 22970    0     0  44492      0 --:--:-- --:--:-- --:--:-- 44492
unpacking source archive /build/589e48ae45c5984d8f1528e1c0b802e9fa137715.tar.gz
error:
       … while calling the 'derivationStrict' builtin

         at /builtin/derivation.nix:9:12: (source not available)

       … while evaluating derivation 'secret-operator'
         whose name attribute is located at /nix/store/zyviry1gxp3y8a82myrcvxyr99p2v64m-nixpkgs/nixpkgs/pkgs/stdenv/generic/make-derivation.nix:300:7

       … while evaluating attribute 'buildInputs' of derivation 'secret-operator'

         at /nix/store/zyviry1gxp3y8a82myrcvxyr99p2v64m-nixpkgs/nixpkgs/pkgs/stdenv/generic/make-derivation.nix:347:7:

          346|       depsHostHost                = lib.elemAt (lib.elemAt dependencies 1) 0;
          347|       buildInputs                 = lib.elemAt (lib.elemAt dependencies 1) 1;
             |       ^
          348|       depsTargetTarget            = lib.elemAt (lib.elemAt dependencies 2) 0;

       error: hash mismatch in fixed-output derivation '/nix/store/60prdsmj7p46jjz25dj2zr5iy7fgk377-source.drv':
         specified: sha256-hLaIY4BE+VIMeKmS3JLOZy87OC2VuQtbX/NCIbQr2p4=
            got:    sha256-ueMqyJJuowrLaqrFrs2YpTZVMvwVxU0Xk/T9Cp9pfe4=
```

nightkr: Did you update the revision without updating the hash?

shell.nix
nightkr: Instead of `import <nixpkgs> {}`, `import (import ./nix/sources.nix).nixpkgs {}`. See https://github.com/nmattia/niv#bootstrapping-a-nix-project.

To an approximation, nix/sources.json is a (JSON representation of) a bunch of `fetchBlah` statements, and `nix/sources.nix` provides an attribute set (dict) that evaluates each of those (so `(import ./nix/sources.nix).nixpkgs` will download the thing described in the `"nixpkgs"` key).

This way, you can `niv add`/`niv update` and get the correct hashes and everything specified automatically.

shell.nix
nightkr: Okay, tested after disabling beku.. it looks like I can get away with disabling `tilt` (definitely grabbed from default.nix, at least I'm not entirely crazy...) and `kind`. Crate2nix also failed inside `--pure` unless I also enabled `cacert`. I don't love not pinning down the others, but I don't think that should block this PR.

shell.nix
NickLarsenNZ: Thanks for your help @nightkr, resolved in 942b0f2dfb7548650ec5c5658d4357a3b25cff54

shell.nix
NickLarsenNZ: Thanks for your help @nightkr, resolved in 942b0f2dfb7548650ec5c5658d4357a3b25cff54

shell.nix
NickLarsenNZ: I would expect an impure shell to be used (so using the certs the developer already trusts). I was just using the pure env to catch all the other dependencies that might not already be on the developer machine.

shell.nix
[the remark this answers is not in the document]
NickLarsenNZ: I took your gist above and massaged it into this shell.nix. I still couldn't get it working without listing dependencies, but have now put them in `nativeBuildInputs`.

See: a13d9e7

NickLarsenNZ:
@dervoeti, if you can please test this for me, that would be great.

1. Clone the repo, checkout this branch `chore/shell.nix`
2. Run a pure nix shell from the root of the repo: `nix shell --pure`
3. Run the test steps:
   ```sh
   cargo clippy --all-targets
   cargo test
   cargo doc --document-private-items
   make regenerate-charts
   make run-dev
   ./scripts/run_tests.sh
   ```

There should be no dependency issues

shell.nix
NickLarsenNZ: @nightkr would you mind trying again when you get the chance?
Some things that have helped others:
- Commit the Cargo.nix and crate-sources.json (which were in .gitignore) 6f472741b850267e659c0b54fa56c85b76317b4e
- Pulling in submodules `git submodule update`

default.nix
```diff
@@ -40,6 +40,7 @@
 , dockerTag ? null
 }:
 rec {
+  inherit cargo sources pkgs;
```
NickLarsenNZ, line 43: @nightkr just realised `default.nix` comes from [operator-templating](https://github.com/stackabletech/operator-templating/blob/main/template/default.nix)

Should I revert this change, and pull in pkgs from sources?

```suggestion
```

shell.nix
```diff
@@ -0,0 +1,55 @@
+let
+  # sources = import ./nix/sources.nix;
+  # pkgs = import sources.nixpkgs {};
+
+  self = import ./. {};
+  inherit (self) sources pkgs;
```
NickLarsenNZ, line 6: continuing from the comment left on `default.nix`...

Should I do this:

```suggestion
  sources = import ./nix/sources.nix;
  pkgs = import sources.nixpkgs {};

  self = import ./. {};
```

default.nix
NickLarsenNZ: Oh no, because shell.nix needs at least `cargo` from `default.nix`.
So should this line go into operator-templating?

dervoeti:
> @dervoeti, if you can please test this for me, that would be great.
> 
> 1. Clone the repo, checkout this branch `chore/shell.nix`
> 2. Run a pure nix shell from the root of the repo: `nix shell --pure`
> 3. Run the test steps:
>    ```shell
>    cargo clippy --all-targets
>    cargo test
>    cargo doc --document-private-items
>    make regenerate-charts
>    make run-dev
>    ./scripts/run_tests.sh
>    ```
> 
> There should be no dependency issues

Works as expected on NixOS 23.11 with the recent changes.
Only adjustments to your comment:
Run `nix-shell` instead of `nix shell`, and `cacert` has to be present as a package, as the comment in `shell.nix` suggests.

shell.nix
nightkr: The reason that I'm somewhat wary of committing the Cargo.nix is that it's another thing to get out of sync when some people use the Nix workflow and some don't.

NickLarsenNZ: Do you mean as changes are made by non Nix users, the Nix stuff might lag?
I imagine a sprinkle of automation to keep that up to date might help there (I mean, I can easily forget to update it too as a Nix user).

By hiding it away, it just adds a bit more friction for those who do use nix. They have to keep running an extra step, and then each would be potentially out of date with the other. So I do think it is best that it is centralised/shared.

Is this enough of a show stopper for you, or can we do it and deal with sync issues as they come (I'd rather not preempt too much at this juncture)?

If so, could it be a think we put on the experimental board that was recently discussed by the team?

nightkr: I think it's fine to try out.

> By hiding it away, it just adds a bit more friction for those who do use nix.

The thought process at the time was that it was less about "let's add a Nix workflow" and more "let's add the Tilt workflow and that happens to involve Nix".

Now that that situation is changing, I think it's fair to try checking it in and see how it ends up working out.

default.nix
NickLarsenNZ: Done in https://github.com/stackabletech/operator-templating/pull/323

shell.nix
NickLarsenNZ: Done in https://github.com/stackabletech/operator-templating/pull/323

nightkr: Personally I prefer keeping `default.nix` as the canonical source to import from.

NickLarsenNZ: Cool, will mark this as resolved.

shell.nix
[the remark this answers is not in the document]
NickLarsenNZ: @nightkr are you ok with that?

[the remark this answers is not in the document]
nightkr: When I tried to build it now, of the block of pkg-config thru krb5, clang was the only that caused build failures when removed. And that could be removed by copying over the `BINDGEN_EXTRA_CLANG_ARGS` from default.nix.

This diff builds fine for in a `nix-shell --pure`: https://gist.github.com/nightkr/a2afc0e29fe3edc9e113a0753826be94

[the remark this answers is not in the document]
NickLarsenNZ: Thanks for the assistance @nightkr. I have tested what you suggested, replacing the extra deps with an env var:

```nix
BINDGEN_EXTRA_CLANG_ARGS = "-I${pkgs.glibc.dev}/include -I${pkgs.clang}/resource-root/include";
```

See 760669a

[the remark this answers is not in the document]
NickLarsenNZ: I will bring the same change in https://github.com/stackabletech/operator-templating/pull/323

NickLarsenNZ:
Superseded by #373 (though it misses the .envrc.sample)

PR #397, opened 2024-04-08
Default to installing the TLS CA into the operator's target namespace

nightkr:
# Description

See https://github.com/stackabletech/issues/issues/498. OLM manifests are generated from the Helm charts, so fixing it in one should fix it in the other.

Note that this is a breaking change, existing users will need to either `--set secretClasses.tls.caSecretNamespace=default` or copy their CA secret across (`kubectl -n default get secret/secret-provisioner-tls-ca -o json | jq '.metadata.namespace = "stackable-operators"' | kubectl create -f-`).

```[tasklist]
# Author
- [ ] CRD changes approved
- [ ] CRD documentation for all fields, following the [style guide](https://docs.stackable.tech/home/nightly/contributor/docs-style-guide).
```

```[tasklist]
# Reviewer
- [ ] Documentation added or updated. Follows the [style guide](https://docs.stackable.tech/home/nightly/contributor/docs-style-guide).
- [ ] Changelog updated
```

```[tasklist]
# Acceptance
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,13 @@ All notable changes to this project will be documented in this file.
 
 ## [Unreleased]
 
+### Changed
+
+- [BREAKING] The TLS CA Secret is now installed into the Namespace of the operator (typically `stackable-operators`), rather than `default`.
+  - Existing users can either migrate (recommended) by copying the CA into the correct Namespace
+    (`kubectl -n default get secret/secret-provisioner-tls-ca -o json | jq '.metadata.namespace = "stackable-operators"' | kubectl create -f-`),
+    or by setting the `secretClasses.tls.caSecretNamespace` Helm flag (`--set secretClasses.tls.caSecretNamespace=default`).
+
 ## [24.3.0] - 2024-03-20
 
 ### Added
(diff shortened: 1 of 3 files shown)


razvan:
This prompted me to have a look at what we've done with 24.3 and it looks like `default` is[ hard coded there too](https://github.com/redhat-openshift-ecosystem/certified-operators/blob/37db568aba0de800b808394cd99e164f605b3be8/operators/stackable-secret-operator/24.3.0/manifests/secret-operator-manifests.yaml#L50).

Then I generated the manifests from this branch and saw that "default" is also used. This is understandable because there is no namespace set at the time when we generated the manifests.

So I think that this is a good fix for Helm but not for OLM where we probably need to interpolate the namespace from the environment at the installation time (or something like that).

I added a new task to this epic: https://github.com/stackabletech/issues/issues/561

nightkr:
This is still going to leave us in a better spot for OLM too though than we were in before, right? At least it would align with the default settings.

razvan (approved):
lgtm but maybe remove the _Fixes_ reference to the issue before merging so that it's not closed automatically.

PR #398, opened 2024-04-09
docs: update example to match description on the page

NickLarsenNZ:
The docs say:

> 2. This secret should be scoped by the intersection of node, pod, and the service named secret-consumer

But the example yaml only showed the `node` scope.


sbernauer (approved):
Thanks!

Issue #453, opened 2024-06-26
Reuse an existing TLS CA from a previous installation


siegfriedweber:
Until SDP 24.3, the SecretClass `tls` used the `default` namespace to store the TLS CA. In #397, the namespace was changed to the operator's namespace in the Helm installation. This is a breaking change and users migrating from a prior SDP version must either revert this change by setting the Helm value `secretClasses.tls.caSecretNamespace` to `default` or by copying the Secret `secret-provisioner-tls-ca` to the operator's namespace.

This change was already implemented and released in the OLM version 24.4.0 (see stackabletech/stackable-utils#65).

It was planned to just add a note to the changelog. But it turned out, that this manual step is error-prone and should be performed automatically.

One solution is to add a Job to the Helm installation which checks if the secret exists in the `default` namespace but not in the one set in `caSecretNamespace` and copies the secret in this case.

PR #476, opened 2024-07-22
feat: migrate CA secret to new namespace

dervoeti:
# Description

Fixes https://github.com/stackabletech/secret-operator/issues/453 by creating a one-time job that copies the secret over to the new namespace if it does not exist there yet.

```[tasklist]
# Author
```

```[tasklist]
# Reviewer
- [ ] Documentation added or updated. Follows the [style guide](https://docs.stackable.tech/home/nightly/contributor/docs-style-guide).
- [ ] Changelog updated
```

```[tasklist]
# Acceptance
```


deploy/helm/secret-operator/templates/secret_migration_job.yaml
```diff
@@ -0,0 +1,54 @@
...
+            kubectl get secret -n $TARGET_NAMESPACE secret-provisioner-tls-ca
+            exit_status=$?
+            # only continue if secret in target namespace does NOT exist
+            if [ $exit_status -eq 1 ]; then
+              echo "secret does not exist in namespace $TARGET_NAMESPACE"
+              # copy secret from default to {{ .Values.secretClasses.tls.caSecretNamespace | default .Release.Namespace }}
+              kubectl get secret -n $SOURCE_NAMESPACE secret-provisioner-tls-ca -o json | jq 'del(.metadata["namespace","creationTimestamp","resourceVersion","selfLink","uid"])' | kubectl apply -n $TARGET_NAMESPACE -f -
+            fi
+          fi
+      restartPolicy: Never
```
NickLarsenNZ, line 54: @Techassi, curious why pre-commit skipped all the checks (incl trailing newline). Any ideas?

image.

It is not set to ignore in: https://github.com/stackabletech/secret-operator/blob/main/.pre-commit-config.yaml

deploy/helm/secret-operator/templates/secret_migration_job.yaml
```diff
@@ -0,0 +1,55 @@
...
+            echo "secret exists in namespace $SOURCE_NAMESPACE"
+            kubectl get secret -n $TARGET_NAMESPACE secret-provisioner-tls-ca
+            exit_status=$?
+            # only continue if secret in target namespace does NOT exist
+            if [ $exit_status -eq 1 ]; then
+              echo "secret does not exist in namespace $TARGET_NAMESPACE"
+              # copy secret from default to {{ .Values.secretClasses.tls.caSecretNamespace | default .Release.Namespace }}
+              kubectl get secret -n $SOURCE_NAMESPACE secret-provisioner-tls-ca -o json | jq 'del(.metadata["namespace","creationTimestamp","resourceVersion","selfLink","uid"])' | kubectl apply -n $TARGET_NAMESPACE -f -
+            fi
+          fi
```
lfrancke, line 54: Run shellcheck please

deploy/helm/secret-operator/templates/secret_migration_job.yaml
nightkr: Generally, pre-commit will only trigger for modified files. For a full run you need to add `--all-files`.

deploy/helm/secret-operator/templates/secret_migration_job.yaml
```diff
@@ -0,0 +1,55 @@
+---
+apiVersion: batch/v1
```
nightkr, line 2:
```suggestion
# Migrates the TLS CA keypair from the hard-coded default namespace to the operator namespace
# See https://github.com/stackabletech/secret-operator/issues/453
apiVersion: batch/v1
```

deploy/helm/secret-operator/templates/secret_migration_job.yaml
```diff
@@ -0,0 +1,55 @@
...
+      containers:
+      - name: migrate-secret
+        image: "{{ .Values.secretMigrationJob.image.repository }}:1.0.0-stackable{{ .Chart.AppVersion }}"
+        imagePullPolicy: {{ .Values.secretMigrationJob.image.pullPolicy }}
+        resources:
+            {{ .Values.secretMigrationJob.resources | toYaml | nindent 12 }}
+        command: ["bash", "-c"]
+        args:
+        - |
+          set -uo pipefail
```
nightkr, line 38:
```suggestion
          set -euo pipefail
```

deploy/helm/secret-operator/templates/secret_migration_job.yaml
```diff
@@ -0,0 +1,55 @@
...
+          SOURCE_NAMESPACE=default
+          TARGET_NAMESPACE={{ .Values.secretClasses.tls.caSecretNamespace | default .Release.Namespace }}
+
+          kubectl get secret -n $SOURCE_NAMESPACE secret-provisioner-tls-ca
+          exit_status=$?
+          # only continue if secret exists
+          if [ $exit_status -eq 0 ]; then
```
nightkr, line 45:
```suggestion
          # only continue if secret exists
          if source_ca_secret="$(kubectl get secret -n $SOURCE_NAMESPACE secret-provisioner-tls-ca -o json)"; then
```

deploy/helm/secret-operator/templates/secret_migration_job.yaml
```diff
@@ -0,0 +1,55 @@
...
+            echo "secret exists in namespace $SOURCE_NAMESPACE"
+            kubectl get secret -n $TARGET_NAMESPACE secret-provisioner-tls-ca
+            exit_status=$?
+            # only continue if secret in target namespace does NOT exist
+            if [ $exit_status -eq 1 ]; then
```
nightkr, line 50:
```suggestion
            # only continue if secret in target namespace does NOT exist
            if ! kubectl get secret -n $TARGET_NAMESPACE secret-provisioner-tls-ca; then
```

deploy/helm/secret-operator/templates/secret_migration_job.yaml
dervoeti: Done, it had a smaller complaint about something that was irrelevant in that case but I fixed it anyways.

deploy/helm/secret-operator/templates/secret_migration_job.yaml
NickLarsenNZ: The yaml file was updated, so it should have triggered, no?
I would at least expect these hooks to run:
- trim trailing whitespace
- fix end of files
- detect aws credentials
- detect private keys
- yamllint

deploy/helm/secret-operator/templates/secret_migration_job.yaml
```diff
@@ -0,0 +1,55 @@
...
+              echo "secret does not exist in namespace $TARGET_NAMESPACE"
+              # copy secret from default to {{ .Values.secretClasses.tls.caSecretNamespace | default .Release.Namespace }}
+              kubectl get secret -n $SOURCE_NAMESPACE secret-provisioner-tls-ca -o json | jq 'del(.metadata["namespace","creationTimestamp","resourceVersion","selfLink","uid"])' | kubectl apply -n $TARGET_NAMESPACE -f -
```
nightkr, line 53:
```suggestion
              echo "$source_ca_secret" | jq 'del(.metadata["namespace","creationTimestamp","resourceVersion","selfLink","uid"])' | kubectl apply -n $TARGET_NAMESPACE -f -
```

nightkr: It's redundant to fetch the source secret twice.

deploy/helm/secret-operator/templates/secret_migration_job.yaml
nightkr: Presumably `-e` was left out because you wanted to do manual error handling for the gets, but we can do that surgically instead with `if ! <COMMAND>` and `<COMMAND> || <ERROR HANDLER>` instead.

dervoeti: Yep that's correct, nice improvements, thanks! Will apply them.

deploy/helm/secret-operator/templates/secret_migration_job.yaml
Techassi: Yes, I agree that at least _some_ of the checks (likely the ones @NickLarsenNZ mentioned) should have run.

Techassi: I feel like the detection of changed files works differently in CI compared to what we expect. Locally, running `pre-commit` without any args will only check _staged_ files. So I think in the CI, there are no _staged_ files, just files. So we might need `--all-files` to force checking of all files.

NickLarsenNZ: Hmm, but we don't do that on others and it picks them up fine. And we don't want to run on all files because the next person has to fix _everything_ (which should not be a barrier to adding their changes).

Techassi: Seems like the solution is mentioned here: https://pre-commit.com/#usage-in-continuous-integration

We should use `pre-commit run --from-ref origin/HEAD --to-ref HEAD` to only check changed files in CI.

NickLarsenNZ: Good spotting! Will have to add that into operator-templating at some point.

NickLarsenNZ (approved):
Approved based on the `shellcheck` pass, and the described testing (in the team call).

dervoeti:
I now created separate RBAC rules for the Job now instead of using the ones from secret-operator.

Reasoning (besides better security):
The job should run before secret-op starts and the `SecretClass` is created, because otherwise a race condition might occur where secret-op is already started but the `Job` did not finish yet, hence the `Secret` is not yet copied over to the new namespace and secret-op would create a new CA. Since the `Job` needs a `ServiceAccount` to view/edit Secrets and that `ServiceAccount` needs to be created before the installation of secret-op as well, the `ServiceAccount` and other RBAC resources have been created and also annotated with the Hook annotations for Helm.
I set the `hook-delete-policy` to `hook-succeeded` so all the resources from this one time job will automatically be deleted if it succeeds. The `hook-weight` annotations make sure the RBAC resources are created before the `Job`.

Just retested it a couple of times, works as expected.

PR #505, opened 2024-09-25
test(kuttl): Install cert-manager

NickLarsenNZ:
# Description

Fixes the broken cert-manager tests.

This installs cert-manager if it doesn't look to already be installed.

It handles concurrent runs (as long as the same script is copied), and cleanup (basically mimicking pre/post run tasks, which kuttl can't do). It does so by using finalizers. The last job remaining will be responsible for cleaning up.

> [!NOTE]
> This might seem overkill, but could serve as a pattern if needed in other tests.

Unfortunately the script can't be shared between tests, as it won't end up in the work directory - so it would need to be copy/pasted if needed in other tests.

## Testing

I have tested it by making a concurrent test job that finishes sooner than the real test:

```bash
./scripts/run-tests --test cert-manager
```

```
--- PASS: kuttl (165.82s)
    --- PASS: kuttl/harness (0.00s)
        --- PASS: kuttl/harness/cert-manager-concurrent_openshift-false (34.82s)
        --- PASS: kuttl/harness/cert-manager-tls_openshift-false (165.81s)
PASS
```

```[tasklist]
# Author
```

```[tasklist]
# Reviewer
- [ ] Changelog updated
```


NickLarsenNZ:
Still needs tidying up (line lengths, and repeated calls).

razvan (approved):
smart!

PR #547, opened 2025-01-21
fix: For cluster internal scopes also add variant without trailing dot

sbernauer:
# Description

*Please add a description here. This will become the commit message of the merge request later.*

```[tasklist]
# Author
- [ ] CRD changes approved
- [ ] CRD documentation for all fields, following the [style guide](https://docs.stackable.tech/home/nightly/contributor/docs/style-guide).
```

```[tasklist]
# Reviewer
- [ ] Changelog updated
```

```[tasklist]
# Acceptance
```

### DIFF
--- a/deploy/helm/secret-operator/values.yaml
+++ b/deploy/helm/secret-operator/values.yaml
@@ -98,7 +98,7 @@ affinity: {}
 
 # When running on a non-default Kubernetes cluster domain, the cluster domain can be configured here.
 # See the https://docs.stackable.tech/home/stable/guides/kubernetes-cluster-domain guide for details.
-# kubernetesClusterDomain: my-cluster.local.
+# kubernetesClusterDomain: my-cluster.local
 
 # Kubelet dir may vary in environments such as microk8s, see https://github.com/stackabletech/secret-operator/issues/229
 kubeletDir: /var/lib/kubelet
--- a/docs/modules/secret-operator/pages/reference/environment-variables.adoc
+++ b/docs/modules/secret-operator/pages/reference/environment-variables.adoc
@@ -19,7 +19,7 @@ cargo run -- run
 
 == KUBERNETES_CLUSTER_DOMAIN
 
-*Default value*: cluster.local.
+*Default value*: cluster.local
 
 *Required*: false
 
@@ -31,6 +31,6 @@ Please see the documentation xref:guides:kubernetes-cluster-domain.adoc[on confi
 
 [source]
 ----
-export KUBERNETES_CLUSTER_DOMAIN=mycluster.local.
+export KUBERNETES_CLUSTER_DOMAIN=mycluster.local
 cargo run -- run
 ----
--- a/tests/templates/kuttl/kerberos-ad/kinit-client.yaml.j2
+++ b/tests/templates/kuttl/kerberos-ad/kinit-client.yaml.j2
@@ -23,9 +23,9 @@ spec:
               echo kiniting node
               kinit -kt /stackable/krb/keytab -p HTTP/$NODE_NAME
               echo kiniting service
-              kinit -kt /stackable/krb/keytab -p HTTP/krb5-client.$NAMESPACE.svc.cluster.local.
+              kinit -kt /stackable/krb/keytab -p HTTP/krb5-client.$NAMESPACE.svc.cluster.local
               echo kiniting pod
-              kinit -kt /stackable/krb/keytab -p HTTP/$POD_NAME.krb5-client.$NAMESPACE.svc.cluster.local.
+              kinit -kt /stackable/krb/keytab -p HTTP/$POD_NAME.krb5-client.$NAMESPACE.svc.cluster.local
           env:
             - name: KRB5_CONFIG
               value: /stackable/krb/krb5.conf
--- a/tests/templates/kuttl/kerberos/secretclass.yaml
+++ b/tests/templates/kuttl/kerberos/secretclass.yaml
@@ -8,10 +8,10 @@ spec:
   backend:
     kerberosKeytab:
       realmName: CLUSTER.LOCAL
-      kdc: krb5-kdc.$NAMESPACE.svc.cluster.local.
+      kdc: krb5-kdc.$NAMESPACE.svc.cluster.local
       admin:
         mit:
-          kadminServer: krb5-kdc.$NAMESPACE.svc.cluster.local.
+          kadminServer: krb5-kdc.$NAMESPACE.svc.cluster.local
       adminKeytabSecret:
         name: secret-operator-keytab
         namespace: $NAMESPACE


rust/operator-binary/src/backend/mod.rs
```diff
@@ -182,6 +182,14 @@ impl SecretVolumeSelector {
     ) -> Result<Vec<Address>, ScopeAddressesError> {
         use scope_addresses_error::*;
         let cluster_domain = &pod_info.kubernetes_cluster_domain;
+
+        // TODO: Docs!!!
+
+        let mut cluster_domains = vec![cluster_domain.to_string()];
+        if let Some(cluster_domain_without_trailing_dot) = cluster_domain.strip_suffix('.') {
+            cluster_domains.push(cluster_domain_without_trailing_dot.to_owned());
+        }
```
nightkr, line 191: In my testing, Kerberos wants consistency and TLS doesn't really care. Either *should* be helped by doing both.

tests/templates/kuttl/cert-manager-tls/consumer.yaml
```diff
@@ -18,7 +18,7 @@ spec:
               set -euo pipefail
               ls -la /stackable/tls
               cat /stackable/tls/tls.crt | openssl x509 -noout -text
-              cat /stackable/tls/tls.crt | openssl x509 -noout -text | grep "DNS:my-tls-service.$NAMESPACE.svc.cluster.local"
+              cat /stackable/tls/tls.crt | openssl x509 -noout -text | grep -P "DNS:my-tls-service.$NAMESPACE.svc.cluster.local\.?"
```
nightkr, line 21: This will just test that _either_ variant exists.

rust/operator-binary/src/backend/kerberos_keytab.rs
```diff
@@ -174,7 +174,9 @@ udp_preference_limit = 1
 
 [domain_realm]
 cluster.local = {realm_name}
+cluster.local. = {realm_name}
 .cluster.local = {realm_name}
+.cluster.local. = {realm_name}
```
nightkr, line 179: IME, this shouldn't be necessary at all (probably since we set the default realm before). But if we do keep it then we should read the actual cluster domain, not hard-code cluster.local specifically.

nightkr (commented):
We should fix the places that depend on the old value here, not just blindly add both.

tests/templates/kuttl/cert-manager-tls/consumer.yaml
sbernauer: Ah I though this is just a debug statement and not an assertion :see_no_evil: 
d4f4ad7 (and yes, the assert is not optimal but didn't find a nice way in a realistic time frame)

rust/operator-binary/src/backend/kerberos_keytab.rs
sbernauer: Good catch, this PR was mostly about trying out if we can fix the TLS cert problems we have.
Fixed it in 5781e66

sbernauer:
General comment on why we are thinking of adding both (with and wo trailing dot) to scopes - both for TLS and Kerberos:

1. We don't know what users will be entering. Performance sensitive users might choose a trailing dot, users with problems might not be able to add a dot
2. The zookeeper `./scripts/run-tests --test smoke_zookeeper-3.9.2_use-server-tls-true_use-client-auth-tls-true_openshift-false` tests failed with `[ERROR] Could not establish secure connection using client certificates!`
    Everything was configured correctly with trailing dots. We could only get the zk client happy by adding a SAN without the dot
3. Let's imaging a 24.11 users first updates secret-op. Until he is able to update all other product operators everything can potentially break, because every talks `cluster.local`, but secret-op only hands out `cluster.local.`
4. Is there a downside of adding both? - At the first glance this PR seems like a "let's be better safe than sorry"

That being said this is a WIP, I would leave it totally up on @dervoeti and @maltesander to decide how to proceed, as they looked at the issue at the first place. I just happened to bump op-rs and run into a failing test

rust/operator-binary/src/backend/mod.rs
sbernauer: Tried to respond in https://github.com/stackabletech/secret-operator/pull/547#issuecomment-2605005175

nightkr:
> We don't know what users will be entering. Performance sensitive users might choose a trailing dot, users with problems might not be able to add a dot

Wouldn't secret-operator know just as well as any other operator? Are you planning on supporting mixed environments?

> We could only get the zk client happy by adding a SAN without the dot

curl was also happy IME too without the dot, so I guess this is one indication that TLS SANs should never have it.

> Let's imaging a 24.11 users first updates secret-op. Until they are able to update all other product operators everything can potentially break, because every talks cluster.local, but secret-op only hands out cluster.local.

Migration is a fair concern, that's true. We should explicitly document those migration paths in the comments.

> Is there a downside of adding both? - At the first glance this PR seems like a "let's be better safe than sorry"

We should know what credentials we're asking for, and why. Whether they need to be included when provisioning manually, and so on.

It's fine to add things to that list, it just shouldn't be something we do blindly.

dervoeti:
I agree that, in general, we should prefer not adding the hostname without the dot if it's not really necessary / we can work around it.

I'm not sure what exactly the scenario was (@sbernauer and/or @maltesander did the research on this) but I think one reason was that Zookeeper does a reverse DNS lookup on the client IP and complains that the client cert is not valid for the returned hostname (without the trailing dot).

That would be a reason to add the alternative hostname to the SANs. Other ways to solve this are trying to fix this in Zookeeper or maybe explicitly not supporting Zookeeper mTLS if you use a cluster domain with a trailing dot. I'm fine with either solution, adding the alternative hostname to the SAN was probably just the easiest way make it work.

dervoeti:
So, we have to make a decision. I can't really comment on the Kerberos related changes, but as far as I understand it, they are not strictly necessary but would make migration easier. In that case I would be fine with not merging these changes if they are controversial, an easy migration path is nice to have but I think it's okay if we don't have it.
Regarding the TLS related change, I can't see a practical negative effect on security if we add the non-FQDN DNS name to the SANs if a cert for a FQDN DNS name is requested. It would ease the migration path (at least in one direction) and, more importantly, it would fix the Zookeeper mTLS issue.
So I'd be in favor of merging at least the TLS related change.

But I'm also fine with not merging this at all and explicitly listing Zookeeper mTLS as "known not to work with FQDN cluster domains yet". In that case we probably still support many setups with FQDN cluster domains with 25.3, so it's better than before.

Opinions @nightkr @sbernauer @maltesander ?

dervoeti:
> For TLS only the non-FQDN variant seems to matter at all, at least in my testing. We should only keep the non-FQDN variant there.

I'll also do some tests with this later, if it works I'm fine with that solution as well.

maltesander:
> For TLS only the non-FQDN variant seems to matter at all, at least in my testing. We should only keep the non-FQDN variant there.

Yeah we definitly need the non-FQDN in there. That fixed most of the problems i had. IIRC zookeeper required the FQDN in the certificate.

I would punt on Kerberos as well. Main thing is to fix the certs?

dervoeti:
I did some tests with Zookeeper yesterday, including mTLS tests with and without FQDN cluster domains, adding just the non-FQDN hostname to the SANs worked fine. Will do some more testing today with other products.
I only changed one line in secret-op: https://github.com/stackabletech/secret-operator/commit/b5f3d512f63964415e6fd7389150eed19aa930b1

dervoeti:
@maltesander @nightkr @sbernauer I created a PR that only adds the non-FQDN variant to the SANs, works fine for me:
https://github.com/stackabletech/secret-operator/pull/564

PR #564, opened 2025-02-14
feat: only add non-FQDN variant of fully qualified domain names to SANs

dervoeti:
# Description
Replaces https://github.com/stackabletech/secret-operator/pull/547

Extends FQDN support to improve DNS lookup performance, see https://github.com/stackabletech/issues/issues/656

We experienced issues, especially with mTLS, when a client connects to a host via its FQDN (`pod.service.svc.cluster.local.`). Currently, only the FQDN (with the trailing dot) is added to the SANs. With this fix, only the non-FQDN version (without trailing dot) is added. This solves the problems in our tests.
This fix should be backwards compatible and not break anything that works currently, it only improves support for FQDN cluster domains.

I tested the fix with several integration tests (including Zookeeper and Kafka mTLS tests) by setting the env var `KUBERNETES_CLUSTER_DOMAIN` to `cluster.local.` for all operators.

Without this fix, at least these tests failed for me:
`tls_kafka-3.8.1_zookeeper-latest-3.9.2_use-client-tls-true_use-client-auth-tls-true_openshift-false`
and
`smoke_zookeeper-3.9.2_use-server-tls-true_use-client-auth-tls-true_openshift-false`

```[tasklist]
# Author
- [x] CRD changes approved
```

```[tasklist]
# Reviewer
- [ ] Changelog updated
```

```[tasklist]
# Acceptance
```

### DIFF
--- a/docs/modules/secret-operator/pages/scope.adoc
+++ b/docs/modules/secret-operator/pages/scope.adoc
@@ -59,5 +59,5 @@ For example, a TLS certificate provisioned by the xref:secretclass.adoc#backend-
 xref:#node[] and xref:#pod[] would contain the following values in its `subjectAlternateName` (SAN) extension field:
 
 * The node's IP address
-* The node's fully qualified domain name (`my-node.example.com`)
-* The pod's fully qualified domain name (`my-pod.my-service.my-namespace.svc.cluster.local`)
+* The node's fully qualified domain name (`my-node.example.com`, trailing dots are removed)
+* The pod's fully qualified domain name (`my-pod.my-service.my-namespace.svc.cluster.local`, trailing dots are removed)
--- a/docs/modules/secret-operator/pages/scope.adoc
+++ b/docs/modules/secret-operator/pages/scope.adoc
@@ -59,5 +59,5 @@ For example, a TLS certificate provisioned by the xref:secretclass.adoc#backend-
 xref:#node[] and xref:#pod[] would contain the following values in its `subjectAlternateName` (SAN) extension field:
 
 * The node's IP address
-* The node's fully qualified domain name (`my-node.example.com`, trailing dots are removed)
-* The pod's fully qualified domain name (`my-pod.my-service.my-namespace.svc.cluster.local`, trailing dots are removed)
+* The node's fully qualified domain name (`my-node.example.com`, without a trailing dot)
+* The pod's fully qualified domain name (`my-pod.my-service.my-namespace.svc.cluster.local`, without a trailing dot)
--- a/rust/operator-binary/src/backend/mod.rs
+++ b/rust/operator-binary/src/backend/mod.rs
@@ -179,7 +179,8 @@ impl SecretVolumeSelector {
         scope: &scope::SecretScope,
     ) -> Result<Vec<Address>, ScopeAddressesError> {
         use scope_addresses_error::*;
-        let cluster_domain = &pod_info.kubernetes_cluster_domain;
+        // Turn FQDNs into bare domain names by removing the trailing dot
+        let cluster_domain = pod_info.kubernetes_cluster_domain.trim_end_matches(".");
         let namespace = &self.namespace;
         Ok(match scope {
             scope::SecretScope::Node => {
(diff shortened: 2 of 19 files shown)


rust/operator-binary/src/backend/mod.rs
```diff
@@ -179,7 +179,8 @@ impl SecretVolumeSelector {
         scope: &scope::SecretScope,
     ) -> Result<Vec<Address>, ScopeAddressesError> {
         use scope_addresses_error::*;
-        let cluster_domain = &pod_info.kubernetes_cluster_domain;
+        // Turn FQDNs into bare domain names by removing the trailing dot
```
sbernauer, line 182: Just FYI that this will trim *all* trailing dots, not just one. Should make no difference in practice though
```suggestion
        // Turn FQDNs into bare domain names by removing the trailing dots
```

sbernauer (changes requested):
I trust your experiments, so change looks good to me :)
But before we merge I would like to see some other reviewer on this.

1. Can you please add a changelog entry? I think we should mark this as breaking, but note that it should be fine (tm)

Looking at https://github.com/stackabletech/secret-operator/pull/547, I also noticed the following things:

2. Please also update `docs/modules/secret-operator/pages/scope.adoc` (mention the truncation)
3. Normally I would not require you to write tests, on the other hand it would be kind of sad if we would throw away the effort of #547 writing some tests.
   Would you mind copying and adopting them into your PR?

dervoeti:
@sbernauer Thanks for the feedback! I tried to address all your points.

docs/modules/secret-operator/pages/scope.adoc
```diff
@@ -59,5 +59,5 @@ For example, a TLS certificate provisioned by the xref:secretclass.adoc#backend-
 xref:#node[] and xref:#pod[] would contain the following values in its `subjectAlternateName` (SAN) extension field:
 
 * The node's IP address
-* The node's fully qualified domain name (`my-node.example.com`)
-* The pod's fully qualified domain name (`my-pod.my-service.my-namespace.svc.cluster.local`)
+* The node's fully qualified domain name (`my-node.example.com`, trailing dots are removed)
```
sbernauer, line 62: Actually, you did not implement this ;)
```rust
let mut addrs = vec![Address::Dns(pod_info.node_name.clone())];
```

dervoeti: Ah! I somehow thought this was covered by the Listener branch, but we should explicitly trim node names as well, you're right. Fixed.

maltesander:
Yeah i like this as the minimal change for experimental FQDNs. Fixed most of the problems i stumbled over.

dervoeti:
> Yeah i like this as the minimal change for experimental FQDNs. Fixed most of the problems i stumbled over. Edit: I remember ZooKeeper was extra special, did you test with that as well (i guess if you tried Kafka)?

Yes, I tested both ZooKeeper and Kafka

rust/operator-binary/src/backend/mod.rs
```diff
@@ -208,7 +209,13 @@ impl SecretVolumeSelector {
                 .listener_addresses
                 .get(name)
                 .context(NoListenerAddressesSnafu { listener: name })?
-                .to_vec(),
+                .iter()
+                .map(|addr| match addr {
+                    // Turn FQDNs into bare domain names by removing the trailing dots
+                    Address::Dns(dns) => Address::Dns(dns.trim_end_matches(".").to_string()),
+                    _ => addr.clone(),
+                })
```
nightkr, line 217: We could unify the cases by moving this out of the `match` block.

Something like

```rs
let mut addrs = match ...;
for addr in &mut addrs {
  if let Address::Dns(dns) = addr {
    let trimmed = dns.trim_end_matches('.');
    if addr != trimmed {
      *addr = trimmed.to_string();
    }
  }
}
Ok(addrs)
```

nightkr: Also this logic should go in `tls`, we mustn't apply it to Kerberos.

dervoeti: I moved the logic and tested it with the ZooKeeper tests, still works as expected :slightly_smiling_face:

sbernauer (approved):
LGTM

sbernauer (approved):

PR #566, opened 2025-02-28
fix: Set the issuer in generated TLS certificates to the subject of the issuing certificate

siegfriedweber:
# Description

If a root CA is used to sign an intermediate CA, which in turn is then used by the secret operator to sign generated TLS certificates, then the verification of the generated certificates fails:

```
$ openssl verify -trusted root_ca.crt -untrusted intermediate_ca.crt tls.crt
CN=generated certificate for pod
error 20 at 0 depth lookup: unable to get local issuer certificate
error tls.crt: verification failed
```

It fails because the issuer of the certificate is set to the issuer of the intermediate CA instead of its subject. Subjects and issuers must build a chain from the root CA to the end-entity certificate.

After applying this fix, the verification works:

```
$ openssl verify -trusted root_ca.crt -untrusted intermediate_ca.crt tls.crt
tls.crt: OK
```

```[tasklist]
# Author
```

```[tasklist]
# Reviewer
- [x] Changelog updated
```

```[tasklist]
# Acceptance
```


nightkr (approved):
Oops - good catch.

lfrancke:
Could you add a release note snippet for this?

We suggest the form "Previously, <this went wrong>. With this release, <this was fixed in the following way>."

siegfriedweber:
## Release Notes

### Platform improvements

#### Bug fixes

* Previously, TLS certificates generated by the secret operator referenced a wrong issuer if the secret operator used an intermediate CA. With this release, the issuer of the generated TLS certificate is correct and using an intermediate CA in the secret operator works.

Issue #345, opened 2024-01-11
Investigate using openat to lock down file access


nightkr:
https://val.packett.cool/blog/use-openat/

This conflicts somewhat with #325, but we probably wouldn't be able to use that for unprivileged mode so both might end up being useful.

PR #572, opened 2025-03-18
feat: Add format-specific annotations to override secret file names

Techassi:
This PR adds support to customize the secret file names using `secrets.stackable.tech` annotations on the volume. The following attributes were added:

- `secrets.stackable.tech/format.tls-pkcs12.keystore-name`
- `secrets.stackable.tech/format.tls-pkcs12.truststore-name`
- `secrets.stackable.tech/format.tls-pem.cert-name`
- `secrets.stackable.tech/format.tls-pem.key-name`
- `secrets.stackable.tech/format.tls-pem.ca-name`

This came up in demo testing during the 25.3.0 SPD release, see https://github.com/stackabletech/demos/issues/157#issuecomment-2729108153.

This PR adds a new test dimension which is used in the `tls` tests. All adjusted tests pass:

```
--- PASS: kuttl (98.75s)
    --- PASS: kuttl/harness (0.00s)
        --- PASS: kuttl/harness/tls_openshift-false_rsa-key-length-3072_custom-secret-names-False (18.34s)
        --- PASS: kuttl/harness/tls_openshift-false_rsa-key-length-2048_custom-secret-names-True (7.51s)
        --- PASS: kuttl/harness/tls_openshift-false_rsa-key-length-3072_custom-secret-names-True (16.37s)
        --- PASS: kuttl/harness/tls_openshift-false_rsa-key-length-2048_custom-secret-names-False (8.31s)
        --- PASS: kuttl/harness/cert-manager-tls_openshift-false (98.73s)
PASS
```

### DIFF
--- a/rust/operator-binary/src/backend/mod.rs
+++ b/rust/operator-binary/src/backend/mod.rs
@@ -25,7 +25,13 @@ use stackable_operator::{
 pub use tls::TlsGenerate;
 
 use self::pod_info::SchedulingPodInfo;
-use crate::format::{SecretData, SecretFormat};
+use crate::format::{
+    well_known::{
+        FILE_PEM_CERT_CA, FILE_PEM_CERT_CERT, FILE_PEM_CERT_KEY, FILE_PKCS12_CERT_KEYSTORE,
+        FILE_PKCS12_CERT_TRUSTSTORE,
+    },
+    SecretData, SecretFormat,
+};
 
 /// Configuration provided by the `Volume` selecting what secret data should be provided
 ///
@@ -96,6 +102,51 @@ pub struct SecretVolumeSelector {
     )]
     pub compat_tls_pkcs12_password: Option<String>,
 
+    /// An alternative name used for the TLS PKCS#12 keystore file.
+    ///
+    /// Has no effect if the `format` is not `tls-pkcs12`.
+    #[serde(
+        rename = "secrets.stackable.tech/format.tls-pkcs12.keystore-name",
+        default = "default_pkcs12_keystore_name"
+    )]
+    pub tls_pkcs12_keystore_name: String,
+
+    /// An alternative name used for the TLS PKCS#12 keystore file.
+    ///
+    /// Has no effect if the `format` is not `tls-pkcs12`.
+    #[serde(
+        rename = "secrets.stackable.tech/format.tls-pkcs12.truststore-name",
+        default = "default_pkcs12_truststore_name"
+    )]
+    pub tls_pkcs12_truststore_name: String,
+
+    /// An alternative name used for the TLS PEM certificate.
+    ///
+    /// Has no effect if the `format` is not `tls-pem`.
+    #[serde(
+        rename = "secrets.stackable.tech/format.tls-pem.cert-name",
+        default = "default_tls_pem_cert_name"
+    )]
+    pub tls_pem_cert_name: String,
+
+    /// An alternative name used for the TLS PEM certificate key.
+    ///
+    /// Has no effect if the `format` is not `tls-pem`.
+    #[serde(
+        rename = "secrets.stackable.tech/format.tls-pem.key-name",
+        default = "default_tls_pem_key_name"
+    )]
+    pub tls_pem_key_name: String,
+
+    /// An alternative name used for the TLS PEM certificate authority.
+    ///
+    /// Has no effect if the `format` is not `tls-pem`.
+    #[serde(
+        rename = "secrets.stackable.tech/format.tls-pem.ca-name",
+        default = "default_tls_pem_ca_name"
+    )]
+    pub tls_pem_ca_name: String,
+
     /// The TLS cert lifetime (when using the [`tls`] backend).
     /// The format is documented in <https://docs.stackable.tech/home/nightly/concepts/duration>.
     #[serde(
@@ -162,6 +213,26 @@ fn default_cert_lifetime() -> Duration {
 
 fn default_cert_jitter_factor() -> f64 {
     tls::DEFAULT_CERT_JITTER_FACTOR
+}
+
+fn default_pkcs12_keystore_name() -> String {
+    FILE_PKCS12_CERT_KEYSTORE.to_owned()
+}
+
+fn default_pkcs12_truststore_name() -> String {
+    FILE_PKCS12_CERT_TRUSTSTORE.to_owned()
+}
+
+fn default_tls_pem_cert_name() -> String {
+    FILE_PEM_CERT_CERT.to_owned()
+}
+
+fn default_tls_pem_key_name() -> String {
+    FILE_PEM_CERT_KEY.to_owned()
+}
+
+fn default_tls_pem_ca_name() -> String {
+    FILE_PEM_CERT_CA.to_owned()
 }
 
 #[derive(Snafu, Debug)]
--- a/rust/operator-binary/src/backend/mod.rs
+++ b/rust/operator-binary/src/backend/mod.rs
@@ -297,3 +297,55 @@ impl SecretBackendError for Infallible {
         match *self {}
     }
 }
+
+#[cfg(test)]
+mod tests {
+    use std::collections::HashMap;
+
+    use serde::de::{value::MapDeserializer, IntoDeserializer};
+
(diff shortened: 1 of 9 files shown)


rust/operator-binary/src/csi_server/node.rs
```diff
@@ -223,10 +244,26 @@ impl SecretProvisionerNode {
...
             .context(publish_error::FormatDataSnafu)?
         {
             let item_path = target_path.join(k);
+
+            // Prevent unwanted path traversals by first canonicalizing the final
+            // path and then validating that the path starts with the base we
+            // expect.
+            let item_path = item_path
+                .canonicalize()
+                .context(publish_error::CanonicalizePathSnafu { path: &item_path })?;
```
nightkr, line 257: This looks like more fodder for "we really should just be doing #345 instead (with NOFOLLOW)".

rust/operator-binary/src/backend/mod.rs
```diff
@@ -96,6 +102,51 @@ pub struct SecretVolumeSelector {
...
+    pub tls_pem_key_name: String,
+
+    /// An alternative name used for the TLS PEM certificate authority.
+    ///
+    /// Has no effect if the `format` is not `tls-pem`.
+    #[serde(
+        rename = "secrets.stackable.tech/format.tls-pem.ca-name",
+        default = "default_tls_pem_ca_name"
+    )]
+    pub tls_pem_ca_name: String,
```
nightkr, line 148: We could `#[serde(flatten)]` in `NamingOptions` instead to avoid this duplication.

rust/operator-binary/src/format/well_known.rs
```diff
@@ -123,6 +123,18 @@ pub struct CompatibilityOptions {
     pub tls_pkcs12_password: Option<String>,
 }
 
+/// Options to customize the well-known format file names.
+///
+/// The fields will either contain the default value or the custom user-provided one. This is also
+/// the reason why the fields are not wrapped in [`Option`].
```
nightkr, line 129: I'd rather provide this defaulting on `NamingOptions` itself than refer to some nebulous  defaulting elsewhere.

rust/operator-binary/src/format/well_known.rs
```diff
@@ -46,24 +46,24 @@ pub enum WellKnownSecretData {
...
             ]
             .into(),
             WellKnownSecretData::TlsPkcs12(TlsPkcs12 {
                 keystore,
                 truststore,
             }) => [
-                (FILE_PKCS12_CERT_KEYSTORE.to_string(), keystore),
-                (FILE_PKCS12_CERT_TRUSTSTORE.to_string(), truststore),
+                (names.tls_pkcs12_keystore_name.to_string(), keystore),
+                (names.tls_pkcs12_truststore_name.to_string(), truststore),
```
nightkr, line 66: We can avoid this cloning if we move the `NamingOptions` instead - we shouldn't need it after this anwyay.

Techassi: Done in 83fea67

rust/operator-binary/src/backend/mod.rs
Techassi: Done in 83fea67.

What do you think about using `#[serde(flatten)]` for `CompatibilityOptions` as well?

nightkr: That probably makes sense, yeah.

rust/operator-binary/src/csi_server/node.rs
nightkr: I think I'm okay with letting this in for 25.3.0 since the same problem already exists for k8sSearch, so it's not technically a regression...

rust/operator-binary/src/backend/mod.rs
Techassi: Done in 94e78c5.

I also removed the `#[serde(deserialize_with = "SecretVolumeSelector::deserialize_some")]` argument because I _think_ it is not needed. I might be wrong tho.

nightkr: Please test both with/without the attribute before removing it - from what I remember, not having it caused some weird deserialization failures.

rust/operator-binary/src/csi_server/node.rs
Techassi: Alright, your call. I'm happy to move forward as is and look into `openat` after the release is done.

rust/operator-binary/src/backend/mod.rs
Techassi: Alright, good to know. I will add a bunch of unit tests for it.

Techassi: Added in 2d0a593.

Not providing the `secrets.stackable.tech/format.compatibility.tls-pkcs12.password` field will succeed deserialization without an error.

Techassi:
The path traversal check will be replaced in a followup PR by a better suited solution which leverages capabilities-based filesystem operations.

rust/operator-binary/src/csi_server/node.rs
```diff
@@ -92,6 +98,12 @@ enum PublishError {
         path: PathBuf,
     },
 
+    #[snafu(display("file path {} must not contain more than one component", path.display()))]
+    InvalidComponentCount { path: PathBuf },
+
+    #[snafu(display("file path {} must not be absolute", path.display()))]
+    InvalidAbsolutePath { path: PathBuf },
```
nightkr, line 105: Nit: We should probably quote these (yes, I realize the other paths in the errors aren't quoted either :/)

```suggestion
    #[snafu(display("file path {path:?} must not contain more than one component"))]
    InvalidComponentCount { path: PathBuf },

    #[snafu(display("file path {path:?} must not be absolute"))]
    InvalidAbsolutePath { path: PathBuf },
```

rust/operator-binary/src/csi_server/node.rs
```diff
@@ -247,23 +241,37 @@ impl SecretProvisionerNode {
...
+                publish_error::InvalidAbsolutePathSnafu { path: &file_path }
             );
 
+            // Ensure that the file path only consists of a single normal
+            // component. This prevents any path traversals up the path using
+            // '..'.
+            ensure!(
+                file_path
+                    .components()
+                    .filter(|c| matches!(c, Component::Normal(_)))
```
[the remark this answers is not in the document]
Techassi: > Actually, with the filter ../foo would be legal (since only the component Normal("foo") remains), which breaks half the point of the check.

True. Didn't realize that. I mostly added the filter to avoid letting through `..` as a path (which would only have a single `ParentDir` component).

I guess we could instead use `components().all(|c| Component::Normal(_))` to ensure all components are `Normal`.

rust/operator-binary/src/csi_server/node.rs
Techassi: Yeah I guess we should. I can also quote the existing ones for consistency.

Techassi: Done in a9be60f6237d1ea61eb3495bb68bc16020d1e136.

rust/operator-binary/src/csi_server/node.rs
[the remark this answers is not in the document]
nightkr: I'm fine either way.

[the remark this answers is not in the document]
Techassi: I went for an earlier check, take a look at 22864d2a45fa65baa4033fcd85b93b27e0e427d7.

nightkr (dismissed):

nightkr:
LGTM, assuming tests pass on your end.

Techassi:
You are right, I will add it right away.

lfrancke:
Can you please link docs and add a release note snippet?

Techassi:
Link to docs: <https://docs.stackable.tech/home/nightly/secret-operator/volume/>

# Release Notes

```adoc
Add support for format-specific annotations to override secret file names.
Names can be customized using secret volume annotations which are listed xref:secret-operator:volume.adoc[in our documentation].
See https://github.com/stackabletech/secret-operator/pull/572[secret-operator#572].
```

PR #573, opened 2025-03-18
feat: Allow the specification of additional trust roots

### LINKED ISSUE #570: Deploy additional CAs with the autoTls backend
# Description

It should be possible to deploy additional CAs with the autoTls backend.

If a company wants to use its own root certificate instead of the auto-generated one, an intermediate certificate can be given to the secret operator. Currently, the secret operator only provides this intermediate certificate in volume mounts. Tools, like kcat, which do not use this certificate as a trust anchor, additionally need the root certificate. Therefore, the SecretClass CRD should be extended to allow defining additional certificates, as follows:

```yaml
---
apiVersion: secrets.stackable.tech/v1alpha1
kind: SecretClass
metadata:
  name: tls-intermediate-cert
spec:
  backend:
    autoTls:
      ca:
        autoGenerate: false
        secret:
          name: intermediate-ca
          namespace: stackable-operators
      additionalTrustRoots:
        - secret:
            name: root-ca
            namespace: stackable-operators
```

The secrets of the additional trust roots can contain multiple entries. The keys are considered to be file names and the values their contents. The extension of a file name defines its content. According to https://en.wikipedia.org/wiki/X.509#Certificate_filename_extensions, the extension `.pem` denotes a stack of base64 encoded DER certificates, the extensions `.cer`, `.cert`, and `.crt` denote either a binary DER certificate or a stack of base64 encoded DER certificates. Other extensions can be supported in the future.

siegfriedweber:
# Description

Allow the specification of additional trust roots in autoTls SecretClasses

Closes #570 

```[tasklist]
# Author
- [x] CRD changes approved
```

```[tasklist]
# Reviewer
- [ ] Changelog updated
```

```[tasklist]
# Acceptance
```

### DIFF
--- a/docs/modules/secret-operator/pages/secretclass.adoc
+++ b/docs/modules/secret-operator/pages/secretclass.adoc
@@ -113,6 +113,10 @@ spec:
         keyGeneration:
           rsa:
             length: 4096
+      additionalTrustRoots:
+        - secret:
+            name: trust-roots
+            namespace: default
       maxCertificateLifetime: 15d # optional
 ----
 
@@ -125,6 +129,13 @@ spec:
 `autoTls.ca.keyGeneration`:: Configures how keys should be generated.
 `autoTls.ca.keyGeneration.rsa`:: Declares that keys should be generated using the RSA algorithm.
 `autoTls.ca.keyGeneration.rsa.length`:: The amount of bits used for generating the RSA key pair. Currently, `2048`, `3072` and `4096` are supported. Defaults to `2048` bits.
+`additionalTrustRoots`:: Configures additional trust roots which are added to the CA files or truststores in the provisioned volume mounts.
+`additionalTrustRoots.secret`::
+    Reference (`name` and `namespace`) to a K8s `Secret` object where the trusted certificates are stored.
+    The extension of a key defines its content:
+    * `.der` denotes a a binary DER certificates.
+    * `.pem` denotes a stack of base64 encoded DER certificates.
+    * `.cer`, `.cert` and `.crt` denote either a binary DER certificate or a stack of base64 encoded DER certificates.
 `autoTls.maxCertificateLifetime`:: Maximum lifetime the created certificates are allowed to have. In case consumers request a longer lifetime than allowed by this setting, the lifetime will be the minimum of both.
 
 [#backend-certmanager]
--- a/docs/modules/secret-operator/pages/secretclass.adoc
+++ b/docs/modules/secret-operator/pages/secretclass.adoc
@@ -133,8 +133,8 @@ spec:
 `additionalTrustRoots.secret`::
     Reference (`name` and `namespace`) to a K8s `Secret` object where the trusted certificates are stored.
     The extension of a key defines its content:
+    * `.crt` denotes a stack of PEM (base64-encoded DER) certificates.
     * `.der` denotes a a binary DER certificates.
-    * `.crt` denotes a stack of base64 encoded DER certificates.
 `autoTls.maxCertificateLifetime`:: Maximum lifetime the created certificates are allowed to have. In case consumers request a longer lifetime than allowed by this setting, the lifetime will be the minimum of both.
 
 [#backend-certmanager]
--- a/docs/modules/secret-operator/pages/secretclass.adoc
+++ b/docs/modules/secret-operator/pages/secretclass.adoc
@@ -134,8 +134,7 @@ spec:
     Reference (`name` and `namespace`) to a K8s `Secret` object where the trusted certificates are stored.
     The extension of a key defines its content:
     * `.der` denotes a a binary DER certificates.
-    * `.pem` denotes a stack of base64 encoded DER certificates.
-    * `.cer`, `.cert` and `.crt` denote either a binary DER certificate or a stack of base64 encoded DER certificates.
+    * `.crt` denotes a stack of base64 encoded DER certificates.
 `autoTls.maxCertificateLifetime`:: Maximum lifetime the created certificates are allowed to have. In case consumers request a longer lifetime than allowed by this setting, the lifetime will be the minimum of both.
 
 [#backend-certmanager]
--- a/rust/operator-binary/src/backend/tls/ca.rs
+++ b/rust/operator-binary/src/backend/tls/ca.rs
@@ -77,7 +77,7 @@ pub enum Error {
         secret: ObjectRef<Secret>,
     },
 
-    #[snafu(display("unsupported certificate format in key {key:?} of {secret}; supported extensions: .cer, .cert, .crt, .pem"))]
+    #[snafu(display("unsupported certificate format in key {key:?} of {secret}; supported extensions: .cer, .cert, .crt, .der, .pem"))]
     UnsupportedCertificateFormat {
         key: String,
         secret: ObjectRef<Secret>,
@@ -484,6 +484,7 @@ impl Manager {
             let extension = Path::new(key).extension().and_then(OsStr::to_str);
             let certs = match extension {
                 Some("pem") => X509::stack_from_pem(value),
+                Some("der") => X509::from_der(value).map(|cert| vec![cert]),
                 Some("cer") | Some("cert") | Some("crt") => X509::from_der(value)
                     .map(|cert| vec![cert])
                     .or(X509::stack_from_pem(value)),
--- a/rust/operator-binary/src/backend/tls/ca.rs
+++ b/rust/operator-binary/src/backend/tls/ca.rs
@@ -501,7 +501,7 @@ impl Manager {
                 secret: secret_ref,
             })?;
             info!(
-                %certs,
+                ?certs,
                 %secret_ref,
                 %key,
                 "adding certificates from additional trust root",
--- a/rust/operator-binary/src/backend/tls/ca.rs
+++ b/rust/operator-binary/src/backend/tls/ca.rs
@@ -562,7 +562,7 @@ impl Manager {
     ) -> Result<Vec<X509>> {
         let extension = Path::new(key).extension().and_then(OsStr::to_str);
 
-        let certs = match extension {
+        match extension {
             Some("crt") => X509::stack_from_pem(value),
             Some("der") => X509::from_der(value).map(|cert| vec![cert]),
             _ => {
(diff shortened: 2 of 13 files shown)


rust/operator-binary/src/backend/tls/ca.rs
```diff
@@ -441,11 +450,68 @@ impl Manager {
...
+            .await
+            .with_context(|_| FindSecretSnafu { secret: secret_ref })?;
+
+        let secret_data = secret.data.unwrap_or_default();
+        for (key, ByteString(value)) in &secret_data {
+            let extension = Path::new(key).extension().and_then(OsStr::to_str);
+            let certs = match extension {
+                Some("pem") => X509::stack_from_pem(value),
+                Some("der") => X509::from_der(value).map(|cert| vec![cert]),
+                Some("cer") | Some("cert") | Some("crt") => X509::from_der(value)
```
nightkr, line 488: *Do* we want to support all these variants?

rust/operator-binary/src/backend/tls/ca.rs
```diff
@@ -441,11 +450,68 @@ impl Manager {
...
+        client: &stackable_operator::client::Client,
+        secret_ref: &SecretReference,
+    ) -> Result<Vec<X509>> {
+        let mut certificates = vec![];
+
+        let secrets_api = &client.get_api::<Secret>(&secret_ref.namespace);
+        let secret = secrets_api
+            .get(&secret_ref.name)
+            .await
+            .with_context(|_| FindSecretSnafu { secret: secret_ref })?;
```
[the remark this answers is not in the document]
siegfriedweber: You are right.

A minor drawback of ConfigMaps is that they distinguish between UTF-8 string data and binary data, which makes them slightly more complicated to use.

The auto-generated certificates and keys of the Secret operator are contained in one Secret, because it is convenient, I guess. They could also be split into a ConfigMap and a Secret. If the user already has a root CA and key in a Secret (I do not claim, that this is a good idea.), then the Secret operator can just use this Secret and ignore the key (e.g. by ignoring entries with the extension `.key`).

I am fine with changing the current implementation from Secret to ConfigMap. But I could also extend the implementation for ConfigMaps:

```yaml
      additionalTrustRoots:
        - secret:
            name: trust-roots
            namespace: default
        - configMap:
            name: trust-roots
            namespace: default
```

What do you prefer?

rust/operator-binary/src/backend/tls/ca.rs
siegfriedweber: These are common extensions, there was no implementation effort, and I guess that this list does not generate maintenance effort in the future.

My thought was, if the user already has the root CA deployed in the cluster, then we increase the chance, that the existing Secret/ConfigMap can be re-used. But on the other side, I violate the YAGNI principle here.

NickLarsenNZ: Just to note (in case it is relevant here), I have seen plenty of PEM certs with the `crt` file extension.

rust/operator-binary/src/backend/tls/ca.rs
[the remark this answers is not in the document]
NickLarsenNZ: FWIW, I think it's ok to use Secret here. It is quite common even though the data itself is not sensitive.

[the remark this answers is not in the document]
nightkr: > A minor drawback of ConfigMaps is that they distinguish between UTF-8 string data and binary data, which makes them slightly more complicated to use.

M'yeah.

> The auto-generated certificates and keys of the Secret operator are contained in one Secret, because it is convenient, I guess. They could also be split into a ConfigMap and a Secret.

They're in one object because they're one cohesive unit to be imported together and updated atomically, and which is fully owned by the one SecretClass.

> If the user already has a root CA and key in a Secret (I do not claim, that this is a good idea.), then the Secret operator can just use this Secret and ignore the key (e.g. by ignoring entries with the extension .key).

Looking beyond 25.3, the way to access just the certificate will be TrustStore, which does create a ConfigMap and also allows formatting the certificates according to your needs (pkcs#12).

> The auto-generated certificates and keys of the Secret operator are contained in one Secret, because it is convenient, I guess.

I think the idiomatic way to do this would be to export the certificate(s) with a TrustStore, and then to import _that_ as an aTR.

The downside there is that that won't make it into 25.3 release. On the other hand, the PR as is doesn't support ignoring key files either.

> I am fine with changing the current implementation from Secret to ConfigMap. But I could also extend the implementation for ConfigMaps:

I think I'd be able to live with saying we support both.

rust/operator-binary/src/backend/tls/ca.rs
nightkr: > My thought was, if the user already has the root CA deployed in the cluster, then we increase the chance, that the existing Secret/ConfigMap can be re-used. But on the other side, I violate the YAGNI principle here.

Mostly I'd just like to promote consistency both internally and with the rest of k8s. But I guess this could also be a question of docs; "You *should* use `ca.crt`, but *can* call it `*.{cer,cert,crt}`.". `.pem`/`.der` make me somewhat worried in terms of overingestion ("it's a zip archive so clearly we must be able to parse it as an OpenDocument"), but on the other hand I guess we'd just fail eagerly either way.

> Just to note (in case it is relevant here), I have seen plenty of PEM certs with the crt file extension.

`.crt` is generally the standard you see in k8s, yeah.

rust/operator-binary/src/backend/tls/ca.rs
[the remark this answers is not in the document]
siegfriedweber: I have now also implemented ConfigMaps.

rust/operator-binary/src/backend/tls/ca.rs
siegfriedweber: I removed all extensions except `.der` where a binary DER certificate is expected and `.crt` where a stack of base64 encoded PEM certificates is expected.

docs/modules/secret-operator/pages/secretclass.adoc
```diff
@@ -134,8 +134,7 @@ spec:
     Reference (`name` and `namespace`) to a K8s `Secret` object where the trusted certificates are stored.
     The extension of a key defines its content:
     * `.der` denotes a a binary DER certificates.
-    * `.pem` denotes a stack of base64 encoded DER certificates.
-    * `.cer`, `.cert` and `.crt` denote either a binary DER certificate or a stack of base64 encoded DER certificates.
+    * `.crt` denotes a stack of base64 encoded DER certificates.
```
nightkr, line 137:
```suggestion
    * `.crt` denotes a stack of PEM (base64-encoded DER) certificates.
```

nightkr: I'd also reorder these to put .crt first.

rust/operator-binary/src/backend/tls/ca.rs
```diff
@@ -481,37 +540,47 @@ impl Manager {
...
+                }
+                .fail();
+            }
+        }
+        .context(LoadCertificateSnafu {
+            key,
+            object: object_ref,
+        })?;
+
+        Ok(certs)
```
nightkr, line 581: nit:
```suggestion
        match extension {
            Some("crt") => X509::stack_from_pem(value),
            Some("der") => X509::from_der(value).map(|cert| vec![cert]),
            _ => {
                return UnsupportedCertificateFormatSnafu {
                    key,
                    object: object_ref,
                }
                .fail();
            }
        }
        .context(LoadCertificateSnafu {
            key,
            object: object_ref,
        })
```

rust/operator-binary/src/crd.rs
```diff
@@ -143,14 +143,26 @@ impl AutoTlsCa {
...
+pub enum AdditionalTrustRoot {
+    /// Reference (name and namespace) to a Kubernetes ConfigMap object where additional
+    /// certificates are stored.
+    /// The extensions of the keys denote its contents: A key suffixed with `.crt` contains a stack
+    /// of base64 encoded DER certificates, a key suffixed with `.der` contains a binary DER
+    /// certificate.
+    ConfigMap {
+        #[serde(flatten)]
+        config_map: ConfigMapReference,
+    },
```
nightkr, line 155: Nit: From what I remember, we could use tuple variants here:

```suggestion
    ConfigMap(ConfigMapReference),
```

But no need to try that right now, it shouldn't affect the API either way.

nightkr (dismissed):

docs/modules/secret-operator/pages/secretclass.adoc
siegfriedweber: done (834f3b2a0a265baf1e9e49b2e43d9379d08b2142)

rust/operator-binary/src/backend/tls/ca.rs
siegfriedweber: done (7214718a08d342d1d00bb208ef628badeae4391c)

rust/operator-binary/src/crd.rs
siegfriedweber: done and successfully tested (a94f08b8f86b46cc137e8266fdfd009ef89d7334)

nightkr (approved):
LGTM in theory, happy to merge if tests pass

docs/modules/secret-operator/pages/secretclass.adoc
```diff
@@ -125,6 +129,12 @@ spec:
 `autoTls.ca.keyGeneration`:: Configures how keys should be generated.
 `autoTls.ca.keyGeneration.rsa`:: Declares that keys should be generated using the RSA algorithm.
 `autoTls.ca.keyGeneration.rsa.length`:: The amount of bits used for generating the RSA key pair. Currently, `2048`, `3072` and `4096` are supported. Defaults to `2048` bits.
+`additionalTrustRoots`:: Configures additional trust roots which are added to the CA files or truststores in the provisioned volume mounts.
+`additionalTrustRoots.secret`::
```
NickLarsenNZ, line 133: I think these are supposed to be:

~~~suggestion
`autoTls.additionalTrustRoots`:: Configures additional trust roots which are added to the CA files or truststores in the provisioned volume mounts.
`autoTls.additionalTrustRoots.secret`::
~~~

siegfriedweber: Thanks! Fixed in #578

PR #597, opened 2025-05-08
 feat: Support putting TrustStore information in Secret 

sbernauer:
## Description

Small follow-up of https://github.com/stackabletech/secret-operator/pull/557

This PR allows the Trustore information (such as the ca.crt) to not only be put in a ConfigMap, but also a Secret.
This was requested by a customer, they mentioned it is required by OpenShift routes, but didn't check that.

The first commit is only test refactoring, I'd suggest reviewing it separately.

## CRD change

```yaml
apiVersion: secrets.stackable.tech/v1alpha1
kind: TrustStore
spec:
  secretClassName: tls
  format: tls-pem

  # New field targetKind:
  targetKind: ConfigMap # or Secret, defaults to ConfigMap
```

### Actual schema change

```yaml
                targetKind:
                  default: ConfigMap
                  description: |-
                    Which Kubernetes resource should be used to output the requested information to.
                    The trust information (such as a `ca.crt`) can be considered public information, so we put it in a `ConfigMap` by default. However, some tools (such as OpenShift routes) require it to be placed in a `Secret`, so we also support that.
                    Can be either `ConfigMap` or `Secret`, defaults to `ConfigMap`.
                  enum:
                    - Secret
                    - ConfigMap
                  type: string
```

### Author

- [x] CRD changes approved

### Reviewer

- [ ] Changelog updated

### Acceptance


nightkr:
Hm, [`Route.spec.tls.destinationCACertificate`](https://docs.redhat.com/en/documentation/openshift_container_platform/4.18/html/network_apis/route-route-openshift-io-v1#spec-tls-3) looks like it wants the certificate inline, so I'm not sure how this would help. 

I also have a vague memory of that we ran into some issues around OpenShift expecting the certificates to use relative hostnames rather than FQDNs...

sbernauer:
I had the same observation, but I think I wouldn't focus to much an OpenShift in particular here.
Customer didn't reply yet, but I guess it's not unrealistic that some tools requires it as Secret, so given the easy implementation I personally would just add it. Would really suck for a customer if he can't use TrustStore because an oversight.

siegfriedweber:
The field `outputResource` could be named `targetResource` (similar to `targetPort` in Services) or `targetKind` because Secret and ConfigMap are kinds not resources (which are instances of kinds).

sbernauer:
Please vote for this CRD change on this comment

sbernauer:
> Hm, [Route.spec.tls.destinationCACertificate](https://docs.redhat.com/en/documentation/openshift_container_platform/4.18/html/network_apis/route-route-openshift-io-v1#spec-tls-3) looks like it wants the certificate inline, so I'm not sure how this would help.

A customer today asked for this, and mentioned
> Could we tell the TrustStore CRD to create a secret instead of a ConfigMap?
> The route config on an Ingress only accepts secret references, not ConfigMaps:
> https://docs.redhat.com/en/documentation/openshift_container_platform/4.14/html/networking/configuring-routes#creating-re-encrypt-route-with-custom-certificate_route-configuration

this should clarify that OpenShift expects it as a Secret

deploy/helm/secret-operator/crds/crds.yaml
```diff
@@ -395,6 +395,18 @@ spec:
                 secretClassName:
                   description: The name of the SecretClass that the request concerns.
                   type: string
+                targetKind:
+                  default: ConfigMap
```
NickLarsenNZ, line 399: I missed the boat on this, but I think `Secret` should be the default.
I assume ConfigMap was chosen as a default for good reason, so please continue.

NickLarsenNZ: Oh I guess because that was the existing behaviour.

NickLarsenNZ (dismissed):
LGTM. Do we need doc updates?

deploy/helm/secret-operator/crds/crds.yaml
sbernauer: Correct, existing behavior is a ConfigMap. If we would default to Secret it's breaking ;)
Also I recall someone in a discussion saying that we should try to be transparent that this is *not* confidential data.

docs/modules/secret-operator/pages/truststore.adoc
```diff
@@ -12,10 +12,12 @@ A TrustStore looks like this:
 include::example$truststore-tls.yaml[]
 ----
 <1> Also used to name the created ConfigMap
-<2> The name of the xref:secretclass.adoc[]
-<3> The requested xref:secretclass.adoc#format[format]
+<2> Mandatory name of the xref:secretclass.adoc[]
+<3> Optional requested xref:secretclass.adoc#format[format]
+<4> Option Kubernetes resource, which should be used to output the requested information to.
```
NickLarsenNZ, line 17:
```suggestion
<4> Optional Kubernetes resource kind, which should be used to output the requested information to.
```

NickLarsenNZ (commented):
Thanks for the docs. Just one correction+suggestion

NickLarsenNZ (approved):
LGTM

lfrancke:
Could you add a release notes snippet and do we have docs for this?

PR #602, opened 2025-05-21
Added comments based on my current understanding of the orchestration code for obtaining keytabs from AD

soenkeliebau:
## Description

I read a small part of the code that is used to generate keytabs from AD and noted down my understanding in comments.
Not sure if correct or helpful to actually merge, but by opening a PR I hope to have a place for discussion.

### Author


rust/krb5-provision-keytab/src/lib.rs
```diff
@@ -71,25 +71,60 @@ pub async fn provision_keytab(krb5_config_path: &Path, req: &Request) -> Result<
...
         .context(SpawnProvisionerSnafu)?;
-    let mut stdin = child.stdin.take().unwrap();
+
+    // Get a `ChildStdin` object for the spawned process and write the serialized request
+    // for a Principal into it in order for the child process to deserialize it and
+    // process the request
+    let mut stdin = child
+        .stdin
+        .take()
+        .expect("Failed to read from stdin of stackable-krb5-provision-keytab script! ");
```
sbernauer, line 99: I *think* this is a bit more accurate, but not sure
```suggestion
        .expect("Failed to read from stdin of stackable-krb5-provision-keytab command! ");
```

sbernauer (approved):

rust/krb5-provision-keytab/src/lib.rs
nightkr: I must've missed that.. it's not even about _reading_, it's about taking ownership of the pipe.

sbernauer (approved):

Issue #627, opened 2025-07-30
Mark samAccountName generation as non-experimental


lfrancke:
In SDP 24.11 we [introduced](https://github.com/stackabletech/secret-operator/pull/454) the possibility to generate samAccountNames automatically
It's been used since then and we haven't seen any problems. Therefore we'd like to mark it as not-experimental anymore.

- [x] Update [documentation](experimentalGenerateSamAccountName)
- [x] Update CRD to not have the `experimental` prefix anymore - potentially use CRD versioning for this?
- [x] Add release notes and mention breaking change

lfrancke:
We'll do https://github.com/stackabletech/secret-operator/issues/627 first to lay the groundwork. Once that is done this one should be simple.

Techassi:
We decided to do #627 first (contrary to the comment above), because the change is slightly easier.

PR #634, opened 2025-08-22
feat: Add v1alpha2 for SecretClass, rename experimentalGenerateSamAccountName

sbernauer:
Part of https://github.com/stackabletech/issues/issues/642

> [!NOTE] 
> Load-balancing over the secret-op DaemonSet Pods did not work on clusters with multiple nodes, see https://github.com/stackabletech/secret-operator/pull/634#issuecomment-3415867736. This has been addressed in #645 and #646.

This PR marks `samAccountName` as non-experimental as per #627. It is the first use of the `stackable-versioned` macro as well as the conversion webhook machinery. See https://github.com/stackabletech/secret-operator/pull/634#issuecomment-3415858644 for the release note.

> [!NOTE] 
> This PR makes changes to things that are templated by operator-templating! We are aware of this and will have a diff until we roll this out to all operators and update operator-templating. secret-operator is our guinea-pig

### DIFF
--- a/deploy/helm/secret-operator/templates/roles.yaml
+++ b/deploy/helm/secret-operator/templates/roles.yaml
@@ -45,11 +45,13 @@ metadata:
 rules:
   # Required to maintain the CRD. The operator needs to do this, as it needs to enter e.g. it's
   # generated certificate in the conversion webhook.
+  {{- if .Values.maintenance.customResourceDefinitions.maintain }}
   - apiGroups: [apiextensions.k8s.io]
     resources: [customresourcedefinitions]
     verbs:
       - create
       - patch
+  {{- end }}
   - apiGroups:
       - ""
     resources:


_test.yaml
sbernauer, line 1: Reminder to remove before merging

rust/operator-binary/src/webhooks/conversion.rs
```diff
@@ -0,0 +1,129 @@
...
+        field_manager: OPERATOR_NAME.to_owned(),
+        namespace: operator_environment.operator_namespace,
+        service_name: operator_environment.operator_service_name,
+    };
+
+    Ok(ConversionWebhookServer::new(crds_and_handlers, options, client).await?)
+}
+
+// !!! TESTING struct!!!
+// !!! Will be removed later!!!
```
sbernauer, line 43: reminder to remove before merging

crds.yaml
Techassi, line 1: I think we should move this to a new folder instead of writing this to a top-level file. Something like "extra" or "crds" comes to mind.

Techassi: Moved in 2c40510.

Makefile
```diff
@@ -117,9 +116,11 @@ config:
 		cp -r deploy/config-spec/* "deploy/helm/${OPERATOR_NAME}/configs";\
 	fi
 
+# We generate a crds.yaml, so that the effect of code changes are visible.
+# The operator will take care of the CRD rollout itself.
 crds:
-	mkdir -p deploy/helm/"${OPERATOR_NAME}"/crds
-	cargo run --bin stackable-"${OPERATOR_NAME}" -- crd | yq eval '.metadata.annotations["helm.sh/resource-policy"]="keep"' - > "deploy/helm/${OPERATOR_NAME}/crds/crds.yaml"
+	mkdir -p extra
+	cargo run --bin stackable-"${OPERATOR_NAME}" -- crd > extra/crds.yaml
```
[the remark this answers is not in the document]
sbernauer: Added a comment in the PR description about this

deploy/helm/secret-operator/crds/crds.yaml
sbernauer, line 1: @razvan reported
```bash
❯ helm upgrade --install --namespace stackable-operators secret-operator deploy/helm/secret-operator/
Error: unable to build kubernetes objects from release manifest: resource mapping not found for name: "tls" namespace: "" from "": no matches for kind "SecretClass" in version "secrets.stackable.tech/v1alpha1"
ensure CRDs are installed first
```

I guess the easiest solution is to move the SecretClass creation into a Job, WDYT?

razvan: can we have an initial crd deployment done by Helm/OLM as before?

Techassi: I would vote "no". There should be no need for it, as the operator immediately deploys it when starting up for the first time.

Techassi: This moved into a decision: https://github.com/stackabletech/decisions/issues/62.

razvan:
I'll put this here for lack of better place now.

I tried installing this op version with OLM on OpenShift.

OLM of course has the same problem as Helm (see comment above). It tries to create a `tls` secret class but there is no CRD for it.

I manually  added the exported CRDs to OLM, but then the operator refuses to start with this error:

```
secret-operator 2025-09-19T10:47:38.191165Z  INFO reconcile_crds: stackable_webhook::servers::conversion: Reconciling CRDs crds=["secretclasses.secrets.sta
kable.tech", "truststores.secrets.stackable.tech"[]
secret-operator Error: failed to run conversion webhook
secret-operator
secret-operator Caused by:
secret-operator     0: failed to reconcile CRDs
secret-operator     1: failed to update CRD "secretclasses.secrets.stackable.tech"
secret-operator     2: ApiError: Apply failed with 2 conflicts: conflicts with "catalog" using apiextensions.k8s.io/v1:
secret-operator        - .spec.versions
secret-operator        - .spec.conversion.strategy: Conflict (ErrorResponse { status: "Failure", message: "Apply failed with 2 conflicts: conflicts with \"
catalog\" using apiextensions.k8s.io/v1:\n- .spec.versions\n- .spec.conversion.strategy", reason: "Conflict", code: 409 })
secret-operator     3: Apply failed with 2 conflicts: conflicts with "catalog" using apiextensions.k8s.io/v1:
secret-operator        - .spec.versions
secret-operator        - .spec.conversion.strategy: Conflict
stream closed EOF for stackable-operators/secret-operator-daemonset-zvvdt (secret-operator)
```

No idea if this is expected or a bug.

My 2 cents: I don't know why you decided to take CRD management out of the hands of package managers because I don't remember discussing it and I find no explanation. But I think one implication of this is that we now have to _implement Helm ourselves_ by having to ensure object creation order at least.

deploy/helm/secret-operator/crds/crds.yaml
Techassi: This (and the connected decision) was discussed amongst @sbernauer, @NickLarsenNZ, and @razvan and we agreed this is the best way forward.

@razvan confirmed in https://github.com/stackabletech/secret-operator/pull/634#issuecomment-3312563238 that the approach works, by manually creating the default custom resource. In the future, this will be done automatically by the operator.

rust/operator-binary/src/main.rs
```diff
@@ -113,46 +119,119 @@ async fn main() -> anyhow::Result<()> {
...
+
             if csi_endpoint
                 .symlink_metadata()
                 .is_ok_and(|meta| meta.file_type().is_socket())
             {
                 let _ = std::fs::remove_file(&csi_endpoint);
             }
+
+            // NOTE (@Techassi): This could maybe be moved into a setup function again. For now,
+            // it is here.
```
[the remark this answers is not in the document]
Techassi: Done in 2e32cd7.

Cargo.toml
```diff
@@ -23,7 +23,7 @@ const_format = "0.2.34"
 futures = { version = "0.3", features = ["compat"] }
 h2 = "0.4"
 hex = "0.4"
-kube-runtime = { version = "1.0", features = ["unstable-runtime-stream-control"] }
+kube-runtime = { version = "2.0.1", features = ["unstable-runtime-stream-control",] }
```
sbernauer:
```suggestion
kube-runtime = { version = "2.0", features = ["unstable-runtime-stream-control"] }
```

.gitignore
```diff
@@ -13,6 +13,7 @@ result
 image.tar
 
 tilt_options.json
+local_values.yaml
```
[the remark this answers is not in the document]
Techassi: No, that's why it is included here. Developers can choose to override/set Helm values with this file as they see fit. This can for example be used to set the `telemetry.consoleLog` level.

rust/operator-binary/src/main.rs
```diff
@@ -113,46 +123,108 @@ async fn main() -> anyhow::Result<()> {
...
+
+    let deserializer = serde_yaml::Deserializer::from_slice(include_bytes!("secretclass.yaml"));
+    let mut tls_secret_class: v1alpha2::SecretClass =
+        serde_yaml::with::singleton_map_recursive::deserialize(deserializer)
+            .expect("compile-time included secretclass must be valid YAML");
+
+    #[rustfmt::skip]
+    let managed_by = Label::managed_by(OPERATOR_NAME, "secretclass").expect("managed-by label must be valid");
+    let version = Label::version(built_info::PKG_VERSION).expect("version label must be valid");
+    let instance = Label::instance(OPERATOR_NAME).expect("instance label must be valid");
```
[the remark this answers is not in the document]
Techassi: I added these labels, because they were there before as well (see deleted `secretclasses.yaml` and `_helpers.tpl`).

rust/operator-binary/src/main.rs
```diff
@@ -113,46 +123,108 @@ async fn main() -> anyhow::Result<()> {
...
+    let name = Label::name(OPERATOR_NAME).expect("name label must be valid");
+
+    tls_secret_class
+        .add_label(managed_by)
+        .add_label(version)
+        .add_label(instance)
+        .add_label(name)
+        .add_label(Label::stackable_vendor());
+
+    tls_secret_class.metadata.namespace = Some(operator_namespace);
```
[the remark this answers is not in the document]
Techassi: Oh yeah you are right, we currently set the value to the previous default value. I can see what I can do to make it configurable again.

rust/operator-binary/src/secretclass.yaml
```diff
@@ -0,0 +1,13 @@
...
+kind: SecretClass
+metadata:
+  name: tls
+spec:
+  backend:
+    autoTls:
+      ca:
+        secret:
+          name: secret-provisioner-tls-ca
+          namespace: stackable-operator
```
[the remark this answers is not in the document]
Techassi: Ah yeah, good point.

rust/operator-binary/src/main.rs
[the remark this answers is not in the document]
sbernauer: I see, that's a good point.

In there we had
```
app.kubernetes.io/instance: {{ .Release.Name }}
```
and
```
{{- if .Chart.AppVersion }}
app.kubernetes.io/version: {{ .Chart.AppVersion | quote }}
{{- end }}
```

Both semantics have actually changed:
1. instance was the name of the helm release, it's now hard-coded "secret-operator", regardless of the helm release name.
2. version was always kept up-to-date with helm, so always the currently deployed secret-op version, which is fine. This code here (correctly) does not do that, it's going to be a random old version eventually

I only see many questions/problems around them and no benefit TBH

.gitignore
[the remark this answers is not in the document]
Techassi: Take a look at: https://github.com/stackabletech/secret-operator/blob/13c9ef79c91000151946333b06f4ea078f9772d2/Tiltfile#L28-L38

This will also be rolled out via `operator-templating`, see https://github.com/stackabletech/operator-templating/pull/554

rust/operator-binary/src/main.rs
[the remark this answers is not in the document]
Techassi: Fixed in b66f3c3 and 5ec590c.

rust/operator-binary/src/main.rs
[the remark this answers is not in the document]
Techassi: So what do you recommend here going forward? Dropping these labels, or updating them to other values?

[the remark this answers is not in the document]
sbernauer: I personally would drop these two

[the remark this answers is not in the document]
Techassi: Removed in 45c3a17.

Techassi:
## Release Note

```adoc
==== Platform improvements

===== Stackable secret-operator

*Breaking:* The custom `samAccountName` generation is no longer experimental.
To make this possible, the secret-operator is the first Stackable operator which supports CRD versioning.

* In version `v1alpha2` of the SecretClass, the `experimentalGenerateSamAccountName` field was renamed to `generateSamAccountName`.
  See the xref:secret-operator:secretclass.adoc[SecretClass reference] for more details.
* The stored version of SecretClass is `v1alpha2`.
  It is however still possible to apply and retrieve SecretClasses in `v1alpha1`.
  The resources are automatically converted by the operator.
TODO: Mention CRD is deployed and maintained by the operator. TLS certs are auto rotated.
```

Techassi:
@sbernauer just discovered we run into issues when the cluster has more than a single node, because the secret-operator is deployed via a DaemonSet. We at least need to ensure that only one instance of the conversion webhook exists because otherwise the Kubernetes API server is unable to verify the TLS certificate.

We will discuss how to move forward on Monday.

Tiltfile
```diff
@@ -17,28 +17,23 @@ custom_build(
...
+
+k8s_yaml(helm(
+   'deploy/helm/' + operator_name,
+   name=operator_name,
+   namespace="stackable-operators",
+   set=[
+      'image.repository=' + registry + '/' + operator_name,
+   ],
+   values=helm_values
+))
```
NickLarsenNZ: We might be able to reduce this PR now:

https://github.com/stackabletech/operator-templating/pull/554/files

Techassi: The templating PR should be merged first then, so that we can incorporate its changes into this PR.

Techassi: Merged in fe17308.

deploy/helm/secret-operator/templates/roles.yaml
```diff
@@ -115,6 +124,10 @@ rules:
       - secretclasses
       - truststores
     verbs:
+      {{- if .Values.maintenance.customResourceDefinitions.maintain }}
```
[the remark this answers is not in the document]
NickLarsenNZ: What are the consequences of gating vs not gating?

Techassi (approved):

NickLarsenNZ (approved):
LGTM, I watched the tests pass

Techassi:
I also ran the Trino integration tests, they all (but two) passed on the first try. The failing ones passed as well when selectively running them.

_I wanted to paste the results here, but sadly I lost the results in my terminal buffer._

Makefile
NickLarsenNZ: this will be overwritten by templating

PR #653, opened 2025-10-31
fix: Force apply CRDs

Techassi:
With this PR, the CRDs are force-applied to become the sole manager of the CRD and its fields. This resolves conflicts with previous deployments via `helm` or `stackablectl`.

See https://github.com/stackabletech/operator-rs/pull/1116 for the relevant change in `stackable-operator`.

This PR was tested by first installing the secret-operator in version 25.7.0 via `stackablectl`. A version with the fixed CRD maintainer code was deployed via `make run-dev`. The CRDs were correctly updated (versions and conversion).

### DIFF
--- a/default.nix
+++ b/default.nix
@@ -5,7 +5,7 @@
     # on Darwin (macOS), but doesn't seem to actually be necessary beyond
     # production hardening.
     fakeroot =
-        if self.buildPlatform.isDarwin then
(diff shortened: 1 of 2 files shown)


Techassi:
I would say this is an implementation detail.

The overall "CRD versioning" stuff is mentioned in our release notes quite extensively.

PR #707, opened 2026-05-18
fix: Refuse to overwrite foreign objects from TrustStore reconciler

dervoeti:
## Description

TLDR: Write access on `TrustStore` allows write access for `ConfigMap` or `Secret` in the same namespace

The TrustStore reconciler writes its output `ConfigMap` or `Secret` using the `TrustStore` CR's own name and applies it via SSA with `force=true`. Combined with the operator's cluster-wide write permissions on `ConfigMaps` and `Secrets`, this lets any principal who can create a `TrustStore` in a namespace cause the operator to overwrite the contents of any pre-existing same-named `ConfigMap` or `Secret` in that namespace, stealing ownership and attaching a controller `OwnerReference` that links the hijacked object to the `TrustStore` CR for cascade delete.

A good showcase is the auto-published `kube-root-ca.crt` ConfigMap:
A tenant with `create` on `truststores.secrets.stackable.tech` in their namespace can create a TrustStore named `kube-root-ca.crt`, overwrite the CA bundle every pod in the namespace uses to verify in-cluster TLS, and cascade-delete that ConfigMap by deleting the TrustStore. This allows a) MITM attacks (hard because you need to modify DNS resolution of the Pods) and b) potentially crashing existing Pods on restart because `kube-root-ca.crt` is corrupted.

The same applies to any `ConfigMap` or `Secret` in the same namespace whose name the attacker can guess.

### Fix

Before applying, look up the existing target. If the object already exists and is not owned by the current `TrustStore`, refuse the reconciliation.

### Author

- [ ] CRD changes approved
- [ ] CRD documentation for all fields, following the [style guide](https://docs.stackable.tech/home/nightly/contributor/docs/style-guide).
- [ ] Links to generated (nightly) docs added
- [ ] Release note snippet added

### Reviewer

- [ ] Changelog updated

### Acceptance

- [ ] Links to generated (nightly) docs added
- [ ] Release note snippet added
- [ ] Add `type/deprecation` label & add to the [deprecation schedule](https://github.com/orgs/stackabletech/projects/44/views/1)
- [ ] Add `type/experimental` label & add to the [experimental features tracker](https://github.com/orgs/stackabletech/projects/47)

### DIFF
--- a/rust/operator-binary/src/truststore_controller.rs
+++ b/rust/operator-binary/src/truststore_controller.rs
@@ -8,6 +8,7 @@ use stackable_operator::{
     k8s_openapi::{
         ByteString,
         api::core::v1::{ConfigMap, Secret},
+        apimachinery::pkg::apis::meta::v1::OwnerReference,
     },
     kube::{
         Resource,
@@ -220,6 +221,27 @@ pub enum Error {
         source: stackable_operator::client::Error,
         secret: ObjectRef<Secret>,
     },
+
+    #[snafu(display(
+        "failed to look up pre-existing target {object} before applying the TrustStore"
+    ))]
+    GetExistingTarget {
+        source: stackable_operator::client::Error,
+        object: ObjectRef<stackable_operator::kube::api::DynamicObject>,
+    },
+
+    #[snafu(display(
+        "refusing to overwrite pre-existing {object} that is not owned by this TrustStore"
+    ))]
+    RefuseToOverwriteForeignTarget {
+        object: ObjectRef<stackable_operator::kube::api::DynamicObject>,
+    },
+
+    #[snafu(display("TrustStore has no associated UID"))]
+    NoTrustStoreUid,
+
+    #[snafu(display("TrustStore has no associated name"))]
+    NoTrustStoreName,
 }
 
 type Result<T, E = Error> = std::result::Result<T, E>;
@@ -239,8 +261,26 @@ impl ReconcilerError for Error {
             Error::BuildOwnerReference { .. } => None,
             Error::ApplyTrustStoreConfigMap { config_map, .. } => Some(config_map.clone().erase()),
             Error::ApplyTrustStoreSecret { secret, .. } => Some(secret.clone().erase()),
+            Error::GetExistingTarget { object, .. } => Some(object.clone()),
+            Error::RefuseToOverwriteForeignTarget { object } => Some(object.clone()),
+            Error::NoTrustStoreUid => None,
+            Error::NoTrustStoreName => None,
         }
     }
+}
+
+/// Returns `true` if `existing_owners` contain a controller `OwnerReference` that points to the
+/// TrustStore with the given UID. Used to ensure we only overwrite output ConfigMaps and Secrets
+/// that we previously created ourselves; refusing otherwise prevents the TrustStore primitive
+/// from being abused to clobber foreign same-named objects (e.g. `kube-root-ca.crt`) via the
+/// operator's elevated cluster-wide write permissions.
+fn is_owned_by_truststore(existing_owners: &[OwnerReference], truststore_uid: &str) -> bool {
+    existing_owners.iter().any(|owner| {
+        owner.controller == Some(true)
+            && owner.kind == "TrustStore"
+            && owner.api_version.starts_with("secrets.stackable.tech/")
+            && owner.uid == truststore_uid
+    })
 }
 
 struct Ctx {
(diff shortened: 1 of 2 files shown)


rust/operator-binary/src/truststore_controller.rs
```diff
@@ -239,10 +261,28 @@ impl ReconcilerError for Error {
...
 
+/// Returns `true` if `existing_owners` contain a controller `OwnerReference` that points to the
+/// TrustStore with the given UID. Used to ensure we only overwrite output ConfigMaps and Secrets
+/// that we previously created ourselves; refusing otherwise prevents the TrustStore primitive
+/// from being abused to clobber foreign same-named objects (e.g. `kube-root-ca.crt`) via the
+/// operator's elevated cluster-wide write permissions.
+fn is_owned_by_truststore(existing_owners: &[OwnerReference], truststore_uid: &str) -> bool {
+    existing_owners.iter().any(|owner| {
+        owner.controller == Some(true)
+            && owner.kind == "TrustStore"
```
sweb: is it possible to take the String from `v1alpha1::TrustStore`?

rust/operator-binary/src/truststore_controller.rs
```diff
@@ -220,6 +221,27 @@ pub enum Error {
...
+
+    #[snafu(display(
+        "refusing to overwrite pre-existing {object} that is not owned by this TrustStore"
+    ))]
+    RefuseToOverwriteForeignTarget {
+        object: ObjectRef<stackable_operator::kube::api::DynamicObject>,
+    },
+
+    #[snafu(display("TrustStore has no associated UID"))]
+    NoTrustStoreUid,
```
sweb: you could move them up to `NoTrustStoreNamespace` to group them a bit

rust/operator-binary/src/truststore_controller.rs
```diff
@@ -333,6 +401,22 @@ async fn reconcile(
                 data: Some(binary_data),
                 ..Default::default()
             };
+            let existing = ctx
```
sweb: the two cases are very similar - you could introduce a helper function as there is now more overlap.

rust/operator-binary/src/truststore_controller.rs
```diff
@@ -311,6 +351,18 @@ async fn reconcile(
         .context(BuildOwnerReferenceSnafu)?
         .build();
 
+    let truststore_name = truststore
```
sweb, line 387: should we forbid certain names like `kube-root-ca.crt` in all cases?

sweb (dismissed):
lgtm - but I primarily did this to dive into the code base, so grains of salt - I also tried out the review skill on this.

rust/operator-binary/src/truststore_controller.rs
dervoeti: Addressed in https://github.com/stackabletech/secret-operator/pull/707/commits/d39f8c70c91360f80a8e403dc387258ddfbaf92a

rust/operator-binary/src/truststore_controller.rs
dervoeti: Addressed in https://github.com/stackabletech/secret-operator/pull/707/commits/d39f8c70c91360f80a8e403dc387258ddfbaf92a

rust/operator-binary/src/truststore_controller.rs
dervoeti: Addressed in https://github.com/stackabletech/secret-operator/pull/707/commits/d39f8c70c91360f80a8e403dc387258ddfbaf92a

dervoeti: We could stuff more into that function (and rename it) but this way the check itself is an isolated addition to the code

rust/operator-binary/src/truststore_controller.rs
dervoeti: I don't think so, the current check should already prevent that scenario (plus others)

NickLarsenNZ (approved):
This looks ready to merge.

NickLarsenNZ:
Merged cancelled due to a comment from @sbernauer 

> IIRC he wanted to do a little bit more digging around it to see what other places of the SDP are "affected". Or if this is all over the place and we are only fixing the tip of the iceberg. In any case we likely want some docs comparing it to cert-manager and trust-manger etc.

dervoeti:
https://github.com/cert-manager/cert-manager has a similar problem but less severe (keys in Secrets/ConfigMaps are fixed and can't be chosen by an attacker), but https://github.com/cert-manager/trust-manager has the same problem. That does not mean we can't do it better though.

We have discovery ConfigMaps in other operators, but that problem is limited to ConfigMaps and the keys are fixed there as well. This problem in secret-operator is more severe since it allows overwriting ConfigMaps and Secrets and the keys can be chosen freely (at least not the content, but still a more severe problem, see the `kube-root-ca.crt` example in the initial comment).

For Listeners, a user can override/delete any Service in the namespace that he can create Listeners in. That is similar to the problem this PR fixes and could be fixed as well.

I did not find any similar problems.

We have three options:
a) merge this and create a PR for listener-operator as well (two local fixes)
b) put the guard into operator-rs so it does ownership checks in general when patching resources (one global fix)
c) leave everything as is and accept the risk

The real security benefit is probably not too high (having permissions to create TrustStores / Listeners is a requirement to exploit this) but it also should not break anything or make it much worse. I'm leaning towards option a) since it fixes the two worst problems and the blast radius is limited.

@sbernauer what's your opinion on this?

sbernauer (approved):
Sure, I let you decide, ship it :)

lfrancke:
Should this have a release note?

dervoeti:
> Should this have a release note?

I don't think so, this should not affect customers in any way, it just strengthens the basic security of SDP a bit.

PR #711, opened 2026-06-16
docs: document CA-only SecretClass for TLS server verification

lfrancke:
## Description

Sibling to
* https://github.com/stackabletech/documentation/pull/860
* https://github.com/stackabletech/nifi-operator/pull/949

Resulting from the same thread.

Describes how to provide a CA for _verification_ and not for creation of actual certificates.


sbernauer (approved):
I think it would also be great to link to https://github.com/stackabletech/documentation/pull/860

PR #721, opened 2026-07-13
docs: clarify kubelet state dir on Tanzu deployments

sweb:
## Description

Clarifies under which conditions VMware Tanzu deployments require a custom `csiNodeDriver.kubeletDir`

(see https://github.com/stackabletech/listener-operator/pull/403)


NickLarsenNZ (approved):
Thanks

PR #747, opened 2026-08-27
chore: Namespace the CSI Helm defined templates

lfrancke:
## Description

The previous name could collide with the identically named thing in listener-operator. This could cause issues when used in an umbrella Chart.

https://github.com/stackabletech/issues/issues/882


dervoeti (approved):

lfrancke:
Release notes will be on the parent.

PR #755, opened 2026-09-10
Skip retrieving listeners for pods that have a deletionTimestamp to avoid deadlock waiting for secret operator pvcs when the listeners have already been garbage collected

### LINKED ISSUE #720: skip the PodListeners lookup when the served Pod is being deleted
## Problem

The secret-operator needs to read `PodListeners` in order to retrieve additional listener addresses to generate into the SAN field of certificates it creates. It has a long timeout waiting for these objects to appear, when a pod is spawned and requests the listener scope on certificates. 
Especially during namespace deletion (maybe at other times as well) this can become a problem, when the `PodListeners` get deleted and the secret operator waits for them indefinitely. Since this is part of the `NodePublishVolume` step, this results in the Pod being stuck and holding `pvc-protection`.

## Fix

In `NodePublishVolume`, if the `Pod` the volume serves has a `deletionTimestamp`, skip the
`PodListeners` lookup and issue the (throwaway) certificate without listener SANs so the mount can
complete. A Pod that is being deleted does not need listener-addressed certs.

The startup/steady-state path is unchanged: live Pods still wait for their listener addresses, so no
correctness regression and no startup race, the guard only relaxes behaviour for Pods already
marked for deletion.

soenkeliebau:
## Description
fixes https://github.com/stackabletech/secret-operator/issues/720

### Author

- [X] Changes are OpenShift compatible
- [X] Helm chart can be installed and deployed operator works
- [X] Integration tests passed (for non trivial changes)
- [X] Changes need to be "offline" compatible

### Reviewer

- [ ] Changelog updated

### Acceptance

- [ ] Links to generated (nightly) docs added
- [ ] Release note snippet added
- [ ] Add `type/deprecation` label & add to the [deprecation schedule](https://github.com/orgs/stackabletech/projects/44/views/1)
- [ ] Add `type/experimental` label & add to the [experimental features tracker](https://github.com/orgs/stackabletech/projects/47)

### DIFF
--- a/rust/operator-binary/src/csi_server/controller.rs
+++ b/rust/operator-binary/src/csi_server/controller.rs
@@ -240,9 +240,11 @@ impl Controller for SecretProvisionerController {
                     .get::<Pod>(&selector.pod, &selector.namespace)
                     .await
                     .context(GetPodSnafu)?;
-                let pod_info = SchedulingPodInfo::from_pod(&self.client, &pod, &selector.scope)
-                    .await
-                    .context(ParsePodSnafu)?;
+                let is_being_deleted = pod.metadata.deletion_timestamp.is_some();
+                let pod_info =
+                    SchedulingPodInfo::from_pod(&self.client, &pod, &selector.scope, is_being_deleted)
+                        .await
+                        .context(ParsePodSnafu)?;
 
                 let backend = backend::dynamic::from_selector(&self.client, &selector)
                     .await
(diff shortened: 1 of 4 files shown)


rust/operator-binary/src/csi_server/controller.rs
```diff
@@ -240,9 +240,15 @@ impl Controller for SecretProvisionerController {
...
-                    .context(ParsePodSnafu)?;
+                let is_being_deleted = pod.metadata.deletion_timestamp.is_some();
+                let pod_info = SchedulingPodInfo::from_pod(
+                    &self.client,
+                    &pod,
+                    &selector.scope,
+                    is_being_deleted,
+                )
+                .await
+                .context(ParsePodSnafu)?;
```
NickLarsenNZ, line 251: `&pod` is already given to `from_pod` as an argument. So I don't think there is a need to pass an extra argument.

The `deletion_timestamp` check can be done inside that function.

